PR #2, opened 2019-11-26
Add benchmark compare function

klauspost:


cli/cmp.go
```diff
@@ -0,0 +1,145 @@
+/*
+ * Warp (C) 2019- MinIO, Inc.
```
harshavardhana:
```suggestion
 * Warp (C) 2019 MinIO, Inc.
```

pkg/bench/compare.go
```diff
@@ -0,0 +1,173 @@
+package bench
```
harshavardhana: missing copyright

klauspost:
@harshavardhana Updated.

PR #12, opened 2019-12-19
Warp distributed benchmarking

klauspost:
# WARP multi client design

## Client

Clients are started with `warp client`. 

`warp client` Only accepts an optional host/ip to listen on, but otherwise no parameters.

The client will listen on port `7761` for RPC call.

Only one server can be connected at the time.

There will be a version check to ensure that clients are compatible with the server.

Note: The reason clients are the ones accepting connections is to make it considerably easier to run on multiple servers. You can basically start them up, do all your benchmarks with different parameters, etc and simply shut them down when done.

# Server

Any benchmark can be run in server mode. When warp is invoked as a server no actual benchmarking will be done on the server.

The server will coordinate the benchmark runs and make sure they are run correctly.

When the benchmark has finished, the combined benchmark info will be collected, merged and saved/displayed.

Enabling server mode is done by adding `-warp.client=client-{1...10}:7761` or a comma separated list of warp client hosts. If no host port is specified the default is added.

Example: `warp get -duration=10m -warp.client=client-{1...10} -host=minio-server-{1...16} -access-key=minio -secret-key=minio123`.

If a warp server is unable to connect to a client the entire benchmark is aborted.

If the warp server looses connection to a client during a benchmark run an error will be displayed, but the benchmark will continue with the remaining clients.

# TODO

* [x] Abort/cancel functionality if server explicitly disconnnects.
* [x] Move inline functions to functions/methods.
* [x] Documentation.
* [x] Add server profiling.
* [x] Record a client ID.


harshavardhana:
Taking this PR in for some customer testing.

PR #24, opened 2020-01-17
Update docs to use --<option> instead of -<option>

harshavardhana:
and fix some typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ If you are [running TLS](https://docs.min.io/docs/how-to-secure-access-to-minio-
 
 All benchmarks operate concurrently. By default the processor determines the number of operations that will be running concurrently. This can however also be tweaked using the `--concurrent` parameter.
 
-Tweaking concurrency can have an impact on performance, especially if there is latency to the server tested. Most benchmarks will also use different prefixes for each "thread" running.
+Tweaking concurrency can have an impact on performance, especially if latency to the server is tested. Most benchmarks will also use different prefixes for each "thread" running.
 
 By default all benchmarks save all request details to a file named `warp-operation-yyyy-mm-dd[hhmmss]-xxxx.csv.zst`. A custom file name can be specified using the `--benchdata` parameter. The raw data is [zstandard](https://facebook.github.io/zstd/) compressed CSV data.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,40 +4,79 @@ S3 benchmarking tool.
...
 `warp command [options]`
 
 # benchmarks
 
-All benchmarks operate concurrently.
-By default the processor determines the number of operations that will be running concurrently.
-This can however also be tweaked using the `-concurrent` parameter.
+All benchmarks operate concurrently. By default the processor determines the number of operations that will be running concurrently. This can however also be tweaked using the `--concurrent` parameter.
+
+Tweaking concurrency can have an impact on performance, especially if there is latency to the server tested. Most benchmarks will also use different prefixes for each "thread" running.
```
kannappanr:
```suggestion
Tweaking concurrency can have an impact on performance, especially if latency to the server is tested. Most benchmarks will also use different prefixes for each "thread" running.
```

README.md
```diff
@@ -4,40 +4,79 @@ S3 benchmarking tool.
...
+
+By default all benchmarks save all request details to a file named `warp-operation-yyyy-mm-dd[hhmmss]-xxxx.csv.zst`. A custom file name can be specified using the `--benchdata` parameter. The raw data is [zstandard](https://facebook.github.io/zstd/) compressed CSV data.
+
+# distributed benchmarking
+
+It is possible to coordinate several warp instances automatically.
+This can be useful for testing performance of a cluster from several clients at once.
+
+For reliable benchmarks clients should have synchronized clocks.
```
kannappanr:
```suggestion
For reliable benchmarks, clients should have synchronized clocks.
```

README.md
```diff
@@ -4,40 +4,79 @@ S3 benchmarking tool.
...
+Warp check whether clocks are within one second of the server,
```
kannappanr:
```suggestion
Warp checks whether clocks are within one second of the server,
```

README.md
```diff
@@ -4,40 +4,79 @@ S3 benchmarking tool.
...
+but optimally clocks should be synchronized with [NTP](http://www.ntp.org/) or a similar service.
```
kannappanr:
```suggestion
but ideally clocks should be synchronized with [NTP](http://www.ntp.org/) or a similar service.
```

README.md
```diff
@@ -4,40 +4,79 @@ S3 benchmarking tool.
...
+
+```
+warp client [listenaddress:port]
+```
+
+`warp client` Only accepts an optional host/ip to listen on, but otherwise no specific parameters.
+By default warp will listen on `127.0.0.1:7761`.
+
+Only one server can be connected at the time.
+However, when a benchmark is done the client can immediately run another one with different parameters.
```
kannappanr:
```suggestion
However, when a benchmark is done, the client can immediately run another one with different parameters.
```

PR #26, opened 2020-01-18
Add Goreleaser script

klauspost:
I'm unable to test since I don't have access to repo secrets, but it builds locally.

`GORELEASER_GITHUB_TOKEN` needs to be [set up as secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#creating-encrypted-secrets).


harshavardhana:
If the plan is to use GitHub as repository for releases, we would have to remove `warp update` command or fix it to make it work.

klauspost:
It can upload to all kind of places. https://goreleaser.com/customization/

harshavardhana:
But that might not be acceptable @klauspost dl.min.io is a fully lock down server with no ssh access without private keys present only in some replicated USBs :smile: 

I had looked at goreleaser in the past for MinIO but we built something away from that a comprehensive list of release binary https://github.com/miniohq/q this is for minio/mc and SDKs

Not saying we should do the same for warp, as it may be an overkill but if warp wants to provide `warp update` like mechanism we can support it to pull from GitHub properly.

Either we remove `warp update` or fix it.

klauspost:
Yeah. You are right. I have removed the update command. Simplifies things a lot.

PR #30, opened 2020-01-24
Add RPM and DEP packages.

klauspost:


klauspost:
Not so controversial. Merging.

PR #37, opened 2020-01-28
Check commands for extra args.

klauspost:


harshavardhana:
You can use ctx.Args().Present() but it's no big deal.

PR #41, opened 2020-01-29
Add Dockerhub publish

klauspost:
Add versions and dockerhub to release.

Probably needs a key in secrets as well.

Binaries are available as artifacts: https://github.com/minio/warp/actions/runs/32798620


dvaldivia:
LGTM

PR #48, opened 2020-02-02
use string constants as needed, use IEC notation

harshavardhana:


cli/delete.go
```diff
@@ -32,7 +32,7 @@ var (
 		},
 		cli.StringFlag{
 			Name:  "obj.size",
-			Value: "1KB",
+			Value: "1MiB",
```
klauspost: There shouldn't be a need to upload this big objects since it is only for delete tests.

harshavardhana: The reason I did that was for faster uploads @klauspost it's too slow.

klauspost: @harshavardhana How is increasing the size helping that?

harshavardhana:
I changed the PR to not change the default obj.size for DELETE as I saw that indeed 1KiB is faster than 1MiB due to lower concurrency used in staging uploads. 

Instead, I moved warp outputs to IEC notation since we use 1024 as the divisor and also use humanize package to parse inputs for obj.size

harshavardhana:
That was just based on this

PR #50, opened 2020-02-03
Add select benchmarks

harshavardhana:


cli/select.go
```diff
@@ -0,0 +1,110 @@
...
+			Usage: "Number of objects to upload.",
+		},
+		cli.StringFlag{
+			Name:  "obj.size",
+			Value: "10MiB",
+			Usage: "Size of each generated object. Can be a number or 10KiB/MiB/GiB. All sizes are base 2 binary.",
+		},
+		cli.StringFlag{
+			Name:  "query",
+			Value: "select * from s3object",
```
klauspost: I wanted to have it do something "more" for the first version, but let's add it and refine later.

But this is a bit too easy to "cheat", but the complexities easily explode.

Let's keep this for now.

PR #58, opened 2020-02-04
Add server access when running clients

klauspost:
All benchmark modes and `analyze` can now spin up a webserver using the hidden `-serve=:8080` parameter.

API for now: 

* GET `/v1/status` - retrieve status while benchmark is running
* GET `/v1/aggregated?segment=1s` - after benchmark, get aggregated data. Segments are 1s.
* GET `/v1/operations` - Download operations compressed.
* GET `/v1/operations/json` - Download operations as JSON (big!)
* GET `/v1` - health check.
* DELETE `/v1` kill the web server. Benchmark finishes if still running.

Use `--json` to enable JSON output in commandline mode.


dvaldivia:
LGTM +1

klauspost:
@harshavardhana License header added.

PR #76, opened 2020-03-06
do not try to publish automatically

harshavardhana:


klauspost:
Why?

klauspost:
@harshavardhana Why?

harshavardhana:
> @harshavardhana Why?

Because we need a more controlled release - we shouldn't be making releases for each PR. Also, we shall more to signing the binaries, etc we cannot host those keys on github.

klauspost:
>  we shouldn't be making releases for each PR

?? That is just because you are creating them. And yes, it does seem silly that you do a release just to get #77 out, which shouldn't change anything.

There are binaries created for pull requests for the 3 major platforms. You can simply link to those instead. Eg. https://github.com/minio/warp/actions/runs/51121659

>  Also, we shall be to signing the binaries, etc we cannot host those keys on github.

Is that ready?

harshavardhana:
Yes I can get it ready anytime, I don't think we should autopublish. We need to be able to build everything for release on a secured server how we do Minio releases, we will still use goreleaser but it is more tightly controlled.

Which not only is able to build the binary, we can sign the binary with gpg, we have even windows signing for MinIO binaries which needs to be done for warp as well. This signed binary is what should be put into the container just like MinIO.

To do all of this in GitHub actions is not possible and perhaps would make this all insecure.

harshavardhana:
Also NOTE: this change only removes making a release automatically, snapshot per PR will still be uploaded.

PR #80, opened 2020-03-08
deprecate MakeBucket, we shall always use pre-created bucket

harshavardhana:


harshavardhana:
changes moved to rewrite-warp branch

PR #86, opened 2020-03-16
rename SSL -> TLS

harshavardhana:


harshavardhana:
A small change taking it in..

PR #87, opened 2020-03-17
Revert "simplify makeBucket behavior (#77)"

klauspost:
This reverts commit bd5cc3d4502ae0de2bfd731c171a1ff7049389ee.

This way we don't require permission to create it, if it already exists.

Not sure what the purpose of #77 was other than to simplify, if any.

### DIFF
--- a/pkg/bench/benchmark.go
+++ b/pkg/bench/benchmark.go
@@ -84,18 +84,26 @@ func (c *Common) GetCommon() *Common {
 func (c *Common) createEmptyBucket(ctx context.Context) error {
 	cl, done := c.Client()
 	defer done()
+	x, err := cl.BucketExists(c.Bucket)
+	if err != nil {
+		return err
+	}
+	if !x {
+		console.Infof("Creating Bucket %q...\n", c.Bucket)
+		err := cl.MakeBucket(c.Bucket, c.Location)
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
Let's not revert it, there is a new design proposal from @abperiasamy which I open a GitHub issue. 

We have to redo how we do our distributed benchmark setup.

Reverting the code leads to bugs which the original PR tried to fix, I will rebase the PR with the latest changes @abperiasamy asked me to do.

klauspost:
@harshavardhana I am more worried about the time aspect since both of us are rather busy right now.

What was the issue it fixed? I can't find any references.

harshavardhana:
The issue was what I reproduced due to racy nature of previous code, once in a while in distributed setups it would fail the runs due to bucket does not exist errors  Sorry I didn't report an issue at that point instead sent a fix as we were in a hurry.

Reproduced it multiple times at customer setups, right now warp works always - of course needs more permissions for the time being to create a bucket which is okay IMO until we fix it correctly.

I shall open a new issue regarding the distributed setup change which @abperiasamy requested..

klauspost:
Just so I know for the future, what is the racy part?

The clients will all check if the bucket exists, if not try to create it. If an error is returned, we check again if it exists (assuming someone else created it).

It seems to me there must be a backend problem if this is racy?

PR #102, opened 2020-04-30
add --disable-multipart flag to allow testing without multipart

harshavardhana:


harshavardhana:
made the change @klauspost PTAL

PR #104, opened 2020-05-04
Add --md5 option

klauspost:
Bonus changes:
* Compress output a bit better
* Clean up put options
* Move `disable-multipart` to IO options
* Add comment value for CSV input, so we can add stuff like cmdline to output later.

Waiting for https://github.com/minio/minio-go/pull/1283

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/dustin/go-humanize v1.0.0
 	github.com/fatih/color v1.7.0
 	github.com/gorilla/websocket v1.4.1
-	github.com/klauspost/compress v1.10.3
+	github.com/klauspost/compress v1.10.5
 	github.com/minio/cli v1.22.0
 	github.com/minio/mc v0.0.0-20200201185513-ac9b9423973b
 	github.com/minio/minio v0.0.0-20200430053134-c7470e6e6eed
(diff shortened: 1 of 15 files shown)


go.mod
```diff
@@ -16,3 +16,5 @@ require (
 	github.com/posener/complete v1.2.2-0.20190702141536-6ffe496ea953
 	github.com/secure-io/sio-go v0.3.0
 )
+
+replace github.com/minio/minio-go/v6 v6.0.55 => ../minio-go
```
harshavardhana, line 20: can you point to the latest release? @klauspost

klauspost: Yeah, it is broken on purpose until I can verify it.

klauspost:
@harshavardhana It is waiting for the linked PR to be merged/released.

Once it is I will make a comparison to see if we want to override.

harshavardhana:
> @harshavardhana It is waiting for the linked PR to be merged/released.
> 
> Once it is I will make a comparison to see if we want to override.

Linked PR is merged we can get it here @klauspost

go.mod
klauspost: Yes saw it, but not released yet. Not much point in setting it to an inbetween version.

klauspost:
@harshavardhana Yes saw it, but not released yet. Not much point in setting it to an inbetween version.

harshavardhana:
> @harshavardhana Yes saw it, but not released yet. Not much point in setting it to an inbetween version.

Releases v6.0.57 warp can be updated

PR #110, opened 2020-06-03
remove redundant version sub-command

harshavardhana:
use --version instead


harshavardhana:
> I would rather do it the other way around. Keep command, remove `--version`.

We removed in all our other tools like mc and minio - if you need more info like `version` we can enhance the cli package to print a nice information.

PR #119, opened 2020-08-11
update minio-go to latest master

harshavardhana:


harshavardhana:
PTAL @klauspost at the earliest, I need this for a customer deployment tomorrow my time.

pkg/bench/get.go
```diff
@@ -186,8 +186,12 @@ func (g *Get) Start(ctx context.Context, wait chan struct{}) (Operations, error)
 				opts.VersionID = obj.VersionID
 				o, err := client.GetObject(ctx, g.Bucket, obj.Name, opts)
 				if err != nil {
-					console.Errorln("download error:", err)
-					op.Err = err.Error()
+					// In-flight requests may get canceled due to
+					// auto-term or reaching the `--duration` specified
+					if !isErrContext(err) {
+						console.Errorln("download error:", err)
+						op.Err = err.Error()
```
klauspost, line 193: This is wrong. It will be counted as a completed request which will skew numbers.

Switch to a non-terminating context like the others.  I will make a PR.

pkg/bench/mixed.go
```diff
@@ -268,8 +268,12 @@ func (g *Mixed) Start(ctx context.Context, wait chan struct{}) (Operations, erro
 					o, err := client.GetObject(nonTerm, g.Bucket, obj.Name, getOpts)
 					fbr.r = o
 					if err != nil {
-						console.Errorln("download error:", err)
-						op.Err = err.Error()
+						// In-flight requests may get canceled due to
```
klauspost, line 271: No, the nonterminating context is used.

pkg/bench/mixed.go
```diff
@@ -288,8 +301,8 @@ func (g *Mixed) Start(ctx context.Context, wait chan struct{}) (Operations, erro
 						console.Errorln(op.Err)
 					}
 					rcv <- op
-					objDone()
 					clDone()
+					objDone()
```
klauspost, line 305: What is the reason for this change?

harshavardhana: a clDone() should have a objDone otherwise we are leaking something here ?

harshavardhana:
Need minio-go updates as well from this PR for the changes in minio-go

harshavardhana:
I updated the PR title to only include minio-go updates @klauspost PTAL

PR #136, opened 2020-09-21
optimize warp to use different hosts for each conn

harshavardhana:
this is useful when `warp` is going through a RR-DNS
or a DNS which resolves to multiple other hosts

so a `--host` is a DNS and resolves to multiple hosts
warp will automatically/randomly use a different host
for each new connection

### DIFF
--- a/cli/client.go
+++ b/cli/client.go
@@ -18,6 +18,7 @@
 package cli
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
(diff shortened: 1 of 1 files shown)


klauspost (changes requested):
I fail to see how this is much different from the current approach since connections are reused.

The default dialer will try the addresses in order whereas this will pick a random. However once a client has a connection it will keep reusing it.

If a RR-DNS returns the same order of clients it isn't really that doing its job since it should be returning entries in a different order for each. Once a connection is open it will not change unless the server closes the connection.

So I am not sure what case this is fixing.

harshavardhana:
Closing this as this is not useful

PR #144, opened 2020-11-03
always create unique objects by using counters

harshavardhana:
with random probability at 2million+ objects,
able to see conflicts i.e same objects are being
overwritten.

overwrites v/s fresh objects have different behavior
to be consistent with benchmarking results we need
to make sure we always use unique objects


klauspost (approved):
LGTM

PR #151, opened 2020-11-19
Set a default concurrency of 20

klauspost:
Let's make the benchmark machine independent by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,7 @@ This will run the benchmark for up to 5 minutes and print the results.
 
 # Benchmarks
 
-All benchmarks operate concurrently. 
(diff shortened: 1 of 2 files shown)


harshavardhana:
> Let's make the benchmark machine independent by default.

Why change this? Do we want to set this lower from 20 if GOMAXPROCS is less than 20?

klauspost:
@harshavardhana We are unlikely to be CPU limited on the client side. 

It is just "a number" and it should normalize performance more on different hardware. 10 seemed low, 30 seemed high, but open to suggestions.

harshavardhana (approved):

PR #173, opened 2021-03-25
add service file automatically upon install

harshavardhana:


klauspost:
Not really a fan of running it by default. It is not really a safe service to have running and I don't want people accidentally leaving it running to be able to create a DDOS attack.

harshavardhana:
> Not really a fan of running it by default. It is not really a safe service to have running and I don't want people accidentally leaving it running to be able to create a DDOS attack.

It's just a service file its not enabled @klauspost

klauspost (approved):
👍🏼

PR #174, opened 2021-04-08
Fix TiB throughput string

klauspost:
Fix missing division and do division on float data.


harshavardhana (changes requested):
Actually we could have just used `humanize.IBytes(t) + '/s'` ? @klauspost

klauspost:
@harshavardhana Not a fan. I like the exact control of precision/switchover points and we should have the precision when dealing with bytes/s.

For example: `Fastest: 20 MiB/s` vs. `Fastest: 20.3MiB/s`

harshavardhana (approved):

PR #180, opened 2021-05-30
move to madmin-go and pkg/

harshavardhana:


klauspost (approved):
LGTM, just fix the complaints.

PR #181, opened 2021-07-27
simplify warp cleanup stage

harshavardhana:


harshavardhana:
Taking this in since this change actually changes the entire cleanup to be super simple and also doesn't unnecessarily overload the server.

PR #184, opened 2021-08-17
analyze: add 25, 75, 90 and 99th for the TTFB

fatpat:
To be able to analyze the TTFB more deeply, let's add the TTFB 25th, 75th, 90th and 99th.


klauspost (approved):
Nice addition. LGTM.

PR #186, opened 2021-08-18
add --resolve-host flag to force DNS resolution to explode multiple A/AAAA records

fatpat:
when s3.xxxx.org has multiple A/AAAA records, I would want warp to explode it to all the IPs without having to specifying all of them by hand.

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -186,6 +186,11 @@ var ioFlags = []cli.Flag{
 		Value: string(hostSelectTypeWeighed),
 		Usage: fmt.Sprintf("Host selection algorithm. Can be %q or %q", hostSelectTypeWeighed, hostSelectTypeRoundrobin),
 	},
+	cli.BoolFlag{
+		Name:   "resolve-host",
+		Usage:  "Resolve the host(s) ip(s) (including multiple A/AAAA records). This can break SSL certificates, use --insecure if so",
+		Hidden: true,
+	},
 	cli.IntFlag{
 		Name:  "concurrent",
 		Value: 20,
--- a/cli/benchserver.go
+++ b/cli/benchserver.go
@@ -92,7 +92,7 @@ func runServerBenchmark(ctx *cli.Context, b bench.Benchmark) (bool, error) {
 		return false, nil
 	}
 
-	conns := newConnections(parseHosts(ctx.String("warp-client")))
+	conns := newConnections(parseHosts(ctx.String("warp-client"), false))
 	if len(conns.hosts) == 0 {
 		return true, errors.New("no hosts")
 	}
--- a/cli/client.go
+++ b/cli/client.go
@@ -51,7 +51,7 @@ const (
 )
 
 func newClient(ctx *cli.Context) func() (cl *minio.Client, done func()) {
-	hosts := parseHosts(ctx.String("host"))
+	hosts := parseHosts(ctx.String("host"), ctx.Bool("resolve-host"))
 	switch len(hosts) {
 	case 0:
 		fatalIf(probe.NewError(errors.New("no host defined")), "Unable to create MinIO client")
@@ -214,7 +214,7 @@ func clientTransport(ctx *cli.Context) http.RoundTripper {
 }
 
 // parseHosts will parse the host parameter given.
-func parseHosts(h string) []string {
+func parseHosts(h string, resolveDNS bool) []string {
 	hosts := strings.Split(h, ",")
 	var dst []string
 	for _, host := range hosts {
@@ -232,7 +232,31 @@ func parseHosts(h string) []string {
 			dst = append(dst, strings.Join(lbls, ""))
 		}
 	}
-	return dst
+
+	if !resolveDNS {
+		return dst
+	}
+
+	var resolved []string
+	for _, hostport := range dst {
+		host, port, _ := net.SplitHostPort(hostport)
+		if host == "" {
+			host = hostport
+		}
+		ips, err := net.LookupIP(host)
+		if err != nil {
+			fatalIf(probe.NewError(err), "Could not get IPs for "+hostport)
+			log.Fatal(err.Error())
+		}
+		for _, ip := range ips {
+			if port == "" {
+				resolved = append(resolved, ip.String())
+			} else {
+				resolved = append(resolved, ip.String()+":"+port)
+			}
+		}
+	}
+	return resolved
 }
 
 // mustGetSystemCertPool - return system CAs or empty pool in case of error (or windows)
@@ -248,7 +272,7 @@ func mustGetSystemCertPool() *x509.CertPool {
 }
 
 func newAdminClient(ctx *cli.Context) *madmin.AdminClient {
-	hosts := parseHosts(ctx.String("host"))
+	hosts := parseHosts(ctx.String("host"), ctx.Bool("resolve-host"))
 	if len(hosts) == 0 {
 		fatalIf(probe.NewError(errors.New("no host defined")), "Unable to create MinIO admin client")
 	}


cli/flags.go
```diff
@@ -186,6 +186,10 @@ var ioFlags = []cli.Flag{
 		Value: string(hostSelectTypeWeighed),
 		Usage: fmt.Sprintf("Host selection algorithm. Can be %q or %q", hostSelectTypeWeighed, hostSelectTypeRoundrobin),
 	},
+	cli.BoolFlag{
+		Name:   "resolve-host",
+		Usage:  "resolv the host(s) ip(s) (including multiple A/AAAA records). Warning: this could break SSL certificates check, use with --insecure if needed.",
```
klauspost, line 191:
```suggestion
		Usage:  "resolve the host(s) ip(s) (including multiple A/AAAA records). This can break SSL certificates, use --insecure if so",
```

klauspost (changes requested):
Requires formatting and a small help text tweak.

cli/flags.go
fatpat: fixed

fatpat:
> Requires formatting and a small help text tweak.

done and I also hide the `--resolve-host` flag as it is quite specific

harshavardhana (commented):
what is the purpose of this flag?

fatpat:
> what is the purpose of this flag?

lets assume that I want to target `https://s3.xxxx.org` and `s3.xxxx.org` resolves to multiple A/AAAA records (`a.b.c.1`, `a.b.c.2` and `a.b.c.3`). When setting `--host s3.xxxx.org` the resolution of the domain is done by the s3 client and it implies cache. Which means that the resolution will be done once for all the concurrent requests and they will all target the same A/AAAA record. We would maybe expect warp to target each IPs from the records instead of just one.

with the `resolve-host` flag, the resolution is done when parsing the `--host` flag and it's converted to a list. It's like if warp was called with `-host a.b.c.1,a.b.c.2,a.b.c.3`.

When using TLS, changing a DNS to an IP could break certificate verification. In this case the `--insecure` flag must be used as a workaround.

harshavardhana:
> > what is the purpose of this flag?
> 
> lets assume that I want to target `https://s3.xxxx.org` and `s3.xxxx.org` resolves to multiple A/AAAA records (`a.b.c.1`, `a.b.c.2` and `a.b.c.3`). When setting `--host s3.xxxx.org` the resolution of the domain is done by the s3 client and it implies cache. Which means that the resolution will be done once for all the concurrent requests and they will all target the same A/AAAA record. We would maybe expect warp to target each IPs from the records instead of just one.
> 
> with the `resolve-host` flag, the resolution is done when parsing the `--host` flag and it's converted to a list. It's like if warp was called with `-host a.b.c.1,a.b.c.2,a.b.c.3`.
> 
> When using TLS, changing a DNS to an IP could break certificate verification. In this case the `--insecure` flag must be used as a workaround.

In Go net.Dial automatically dials the fastest connection,  have you explored writing this with a custom DialContext instead? @fatpat 

```go
var rng = rand.New(rand.NewSource(time.Now().UTC().UnixNano()))

type dialContext func(ctx context.Context, network, address string) (net.Conn, error)

func newCustomDialContext(dialTimeout, dialKeepAlive time.Duration) dialContext {
        return func(ctx context.Context, network, addr string) (net.Conn, error) {
                dialer := &net.Dialer{
                        Timeout:   dialTimeout,
                        KeepAlive: dialKeepAlive,
                }

                host, port, err := net.SplitHostPort(addr)
                if err != nil {
                        return nil, err
                }

                addrs, err := net.LookupHost(host)
                if err != nil {
                        addrs = []string{host}
                }

                for i := range addrs {
                        addrs[i] = net.JoinHostPort(addrs[i], port)
                }

                return dialer.DialContext(ctx, network, addrs[rng.Intn(len(addrs))])
        }
}
```

fatpat:
> > > what is the purpose of this flag?
> > 
> > 
> > lets assume that I want to target `https://s3.xxxx.org` and `s3.xxxx.org` resolves to multiple A/AAAA records (`a.b.c.1`, `a.b.c.2` and `a.b.c.3`). When setting `--host s3.xxxx.org` the resolution of the domain is done by the s3 client and it implies cache. Which means that the resolution will be done once for all the concurrent requests and they will all target the same A/AAAA record. We would maybe expect warp to target each IPs from the records instead of just one.
> > with the `resolve-host` flag, the resolution is done when parsing the `--host` flag and it's converted to a list. It's like if warp was called with `-host a.b.c.1,a.b.c.2,a.b.c.3`.
> > When using TLS, changing a DNS to an IP could break certificate verification. In this case the `--insecure` flag must be used as a workaround.
> 
> In Go net.Dial automatically dials the fastest connection, have you explored writing this with a custom DialContext instead? @fatpat
> 
> ```go
> var rng = rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
> 
> type dialContext func(ctx context.Context, network, address string) (net.Conn, error)
> 
> func newCustomDialContext(dialTimeout, dialKeepAlive time.Duration) dialContext {
>         return func(ctx context.Context, network, addr string) (net.Conn, error) {
>                 dialer := &net.Dialer{
>                         Timeout:   dialTimeout,
>                         KeepAlive: dialKeepAlive,
>                 }
> 
>                 host, port, err := net.SplitHostPort(addr)
>                 if err != nil {
>                         return nil, err
>                 }
> 
>                 addrs, err := net.LookupHost(host)
>                 if err != nil {
>                         addrs = []string{host}
>                 }
> 
>                 for i := range addrs {
>                         addrs[i] = net.JoinHostPort(addrs[i], port)
>                 }
> 
>                 return dialer.DialContext(ctx, network, addrs[rng.Intn(len(addrs))])
>         }
> }
> ```

@harshavardhana : nope. But I'll definitely look into it. It could be a cleaner solution.

fatpat:
@harshavardhana your code works like a charm. This is much cleaner and it works with tls without the need to add `--insecure` :+1:

klauspost:
@harshavardhana @fatpat With "exploded" hosts, we will get stats for each host and we control the hosts selection, so I think the PR as proposed gives more value than a custom dialer.

harshavardhana:
> @harshavardhana @fatpat With "exploded" hosts, we will get stats for each host and we control the hosts selection, so I think the PR as proposed gives more value than a custom dialer.

yes I realized @klauspost since it randomizes it may hit the same server and the weighted distribution would have an effect. 

Sorry @fatpat we may have to go back. 

We can think about adding weighted distribution directly to Dialer itself if needed but that may be a bit more time consuming implementation and might not be needed at this point in time.

fatpat:
ok, patch has been rolled back. 

I added the `--shuffle-hosts` flag to control whether or not we want to shuffle the host list.

harshavardhana (approved):

cli/flags.go
```diff
@@ -186,6 +186,16 @@ var ioFlags = []cli.Flag{
 		Value: string(hostSelectTypeWeighed),
 		Usage: fmt.Sprintf("Host selection algorithm. Can be %q or %q", hostSelectTypeWeighed, hostSelectTypeRoundrobin),
 	},
+	cli.BoolFlag{
+		Name:   "resolve-host",
+		Usage:  "Resolve the host(s) ip(s) (including multiple A/AAAA records). This can break SSL certificates, use --insecure if so",
+		Hidden: true,
+	},
+	cli.BoolFlag{
+		Name:   "shuffle-hosts",
```
klauspost, line 195: I don't really see the point of this. Shuffling shouldn't really have any practical implications.

I would rather that it is semi-repeatable. IMO the option should just be removed.

fatpat: I removed the shuffle-hosts option

sorry for the long long delay of fixing :)

fatpat:
@klauspost can you review and maybe merge this please ?

klauspost (approved):
LGTM

PR #187, opened 2021-08-18
add --disable-http-keepalive to disable HTTP Keep-Alive (connections reuse)

fatpat:
It could be useful for some specific benchmarks


klauspost:
@fatpat Could you "go fmt" the code?

Considering if it should be a hidden parameter. Seems like a quite specialized test to do.

fatpat:
> @fatpat Could you "go fmt" the code?
> 
> Considering if it should be a hidden parameter. Seems like a quite specialized test to do.

this has been fixed

harshavardhana:
Should we mark the results that are obtained after this command to be specifically called out? in the warp output? - This helps specifically differentiate the results.

fatpat:
> Should we mark the results that are obtained after this command to be specifically called out? in the warp output? - This helps specifically differentiate the results.

it should be the same with all the flags. Other flags are not marked in the results I guess. Maybe just add a header at the beginning of the output reminding the calling arguments ?

harshavardhana:
> > Should we mark the results that are obtained after this command to be specifically called out? in the warp output? - This helps specifically differentiate the results.
> 
> it should be the same with all the flags. Other flags are not marked in the results I guess. Maybe just add a header at the beginning of the output reminding the calling arguments ?

Keep-alive change is a fundamental change to how the HTTP connection happens - other args just API level changes not protocol level.

fatpat:
@harshavardhana how would like the ouput to be ?

klauspost:
@harshavardhana @fatpat The commandline used, except access/secret is at the end of the CSV

image

fatpat:
so what do you want to do about outputting this ?

harshavardhana:
> so what do you want to do about outputting this ?

For now it's fine. I will add one later after some discussion internally.

harshavardhana (approved):

klauspost:
@fatpat It is trivial to extract:

```
λ zstd -q -d -c warp-stat-2021-08-12[190851]-B9Og.csv.zst|tail -n 1
# warp stat --host=127.0.0.1:{9001...9002} --access-key=*REDACTED* --secret-key=*REDACTED* --obj.size=60KB --duration=1m0s
```

PR #197, opened 2021-11-23
Add last request access stats

klauspost:
Similar to stats for first access to an object, provide last object access stats, but only print them in detailed mode.


harshavardhana (commented):
WDYT this is useful for?

klauspost:
@harshavardhana To observe if there are speed changes over time.

For example, in our cases, when adding X versions it is interesting to compare the time to add the first version compared to the last version.

harshavardhana (approved):

Issue #210, opened 2022-04-26
Need support for passing --objects in put workload


vivekanandan-k-rh:
**Description:**

Need support for passing --objects in warp put workload. As of now, support is available for passing --objects only in mixed workloads where the benchmark starts after dumping the number of objects mentioned with --objects. Put workload needs support for passing the number of objects to upload with benchmark happening.

klauspost:
Benchmarks are defined by the time duration you want to run. I don't want a second termination parameter which will just make it confusing.

You can use GET benchmarks with a short duration. This will also record the time the upload took.

PR #211, opened 2022-04-26
Added support for passing --objects in put operation with benchmarking

vivekanandan-k-rh:
Summary:
Added support for passing --objects in put operation with benchmarking

Description:
With the code changes done, the put method can upload a specific number of objects with benchmarking.
Following are the scenarios for passing --objects:
1. when "--objects <object_count>" is passed, the specified number of objects will be uploaded to the bucket
2. when "--objects <object_count>" is not passed, benchmarking will happen till the default duration

### DIFF
--- a/cli/put.go
+++ b/cli/put.go
@@ -26,6 +26,11 @@ import (
 
 var (
 	putFlags = []cli.Flag{
+		cli.IntFlag{
+			Name:  "objects",
+			Value: 0,
+			Usage: "Number of objects to upload.",
+		},
 		cli.StringFlag{
 			Name:  "obj.size",
 			Value: "10MiB",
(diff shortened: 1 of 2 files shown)


vivekanandan-k-rh:
This will resolve the issue #210

klauspost:
As noted in #210 I don't see much reason to add this. Maybe there could be a small case if it was called `max-objects`, but only really if you are afraid to fill up the storage.

Adding @harshavardhana for a second opinion.

PS. There are bugs in the code, but that isn't important right now.

harshavardhana:
> As noted in #210 I don't see much reason to add this. Maybe there could be a small case if it was called `max-objects`, but only really if you are afraid to fill up the storage.

Agreed @klauspost let's not take this as this is not useful

PR #220, opened 2022-07-01
update all deps move to go1.17 as default

harshavardhana:


klauspost (commented):
A few go fmt didn't go through. Otherwise LGTM

PR #222, opened 2022-07-15
Add parallel multipart upload benchmark

klauspost:
Allow upload to the same object from multiple clients.

Individual part uploads are recorded as separate put operations.


cli/multipart.go
```diff
@@ -0,0 +1,138 @@
+/*
+ * Warp (C) 2019-2020 MinIO, Inc.
```
harshavardhana, line 2:
```suggestion
 * Warp (C) 2022 MinIO, Inc.
```

pkg/bench/multipart.go
```diff
@@ -0,0 +1,264 @@
...
+					File:     obj.Name,
+					ObjPerOp: 1,
+					Endpoint: client.EndpointURL().String(),
+				}
+				op.Start = time.Now()
+				// TODO: Hack, we should set 'partNumber' parameter.
+				err := opts.SetRange(int64(part)*obj.Size-obj.Size, int64(part)*obj.Size-1)
+				if err != nil {
+					console.Fatal(err)
+				}
```
harshavardhana, line 229: we do have opts.PartNumber @klauspost

harshavardhana (changes requested):

harshavardhana (approved):
LGTM I shall give this a spin.

PR #234, opened 2022-11-07
Added obj.dist to accept object size distributions during object generation

anishsana:
1. Added vscode directory to gitignore.
2. Added the option `obj.dist` to allow the specification of object size distributions.
3. Modified generator tests to account for the new option.


anishsana:
Opened a PR accidentally. Closing it.

PR #244, opened 2023-02-07
support min/max obj.size to support better ranges

harshavardhana:
--obj.size "1KiB,10MiB"


klauspost:
So it takes a bit. I will have to test how the distribution is after this change.

klauspost:
Replaced by #245

PR #246, opened 2023-02-09
Always print concurrency to output

klauspost:
This is important information.

### DIFF
--- a/cli/analyze.go
+++ b/cli/analyze.go
@@ -326,7 +326,7 @@ func printAnalysis(ctx *cli.Context, o bench.Operations) {
 			if details && ops.Concurrency > 0 {
 				console.Printf("Operation: %v (%d). Ran %v. %vConcurrency: %d.%s\n", typ, ops.N, ran, sz, ops.Concurrency, hostsString)
 			} else {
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM

cli/analyze.go
```diff
@@ -326,7 +326,7 @@ func printAnalysis(ctx *cli.Context, o bench.Operations) {
 			if details && ops.Concurrency > 0 {
 				console.Printf("Operation: %v (%d). Ran %v. %vConcurrency: %d.%s\n", typ, ops.N, ran, sz, ops.Concurrency, hostsString)
 			} else {
-				console.Printf("Operation: %v\n", typ)
+				console.Printf("Operation: %v. Concurrency: %v\n", typ, ops.Concurrency)
```
harshavardhana, line 329: should it be `opts.Concurrency * clients` ?

harshavardhana (changes requested):

cli/analyze.go
klauspost: No, the concurrency is the accumulated across all clients.

harshavardhana (approved):

PR #250, opened 2023-02-27
[get] bench object from listing the bucket

fatpat:
when doing a get bench, during the prepare phase, instead of putting objects in the bucket use the objects already in there. This can be used to bench a specific use case with an already filled in bucket. It can also be used to speed up get tests by always using the same objects and gain the prepare phase.

New options:
* `--get-from-listing-bucket`: to activate the listing bucket (current
  behaviour remains)
* `--listing-bucket-max-objects`: to limit the number of object to use
* `--listing-bucket-prefix`: to use a specific prefix during listing
* `--listing-bucket-no-recursive`: disable recursive listing

### DIFF
--- a/cli/get.go
+++ b/cli/get.go
@@ -25,6 +25,14 @@ import (
 )
 
 var getFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "list-existing",
+		Usage: "Instead of preparing the bench by PUTing some objects, only use objects already in the bucket",
+	},
+	cli.BoolFlag{
+		Name:  "list-flat",
+		Usage: "When using --list-existing, do not use recursive listing",
+	},
 	cli.IntFlag{
 		Name:  "objects",
 		Value: 2500,
@@ -82,6 +90,9 @@ func mainGet(ctx *cli.Context) error {
 		RandomRanges:  ctx.Bool("range"),
 		CreateObjects: ctx.Int("objects"),
 		GetOpts:       minio.GetObjectOptions{ServerSideEncryption: sse},
+		ListExisting:  ctx.Bool("list-existing"),
+		ListFlat:      ctx.Bool("list-flat"),
+		ListPrefix:    ctx.String("prefix"),
 	}
 	return runBench(ctx, &b)
 }
--- a/pkg/bench/get.go
+++ b/pkg/bench/get.go
@@ -38,6 +38,9 @@ type Get struct {
 	Collector     *Collector
 	objects       generator.Objects
 	Versions      int
+	ListExisting  bool
+	ListFlat      bool
+	ListPrefix    string
 
 	// Default Get options.
 	GetOpts minio.GetObjectOptions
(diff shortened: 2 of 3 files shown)


pkg/bench/get.go
```diff
@@ -47,6 +51,56 @@ type Get struct {
 // Prepare will create an empty bucket or delete any content already there
 // and upload a number of objects.
 func (g *Get) Prepare(ctx context.Context) error {
+
+	/*
+	 * Prepare the bench by listing object from the bucket
+	 */
+	if g.GetFromListing {
```
klauspost, line 58: We do not use block comments:
```suggestion
	// Prepare the bench by listing object from the bucket
	if g.GetFromListing {
```

pkg/bench/get.go
```diff
@@ -47,6 +51,56 @@ type Get struct {
...
+		cl, done := g.Client()
+
+		// ensure the bucket exist
+		found, err := cl.BucketExists(ctx, g.Bucket)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return(fmt.Errorf("bucket %s does not exist and --get-from-bucket-listing has been set", g.Bucket))
```
klauspost, line 67:
```suggestion
			return(fmt.Errorf("bucket %s does not exist and --list-existing has been set", g.Bucket))
```

pkg/bench/get.go
```diff
@@ -33,11 +33,15 @@ import (
...
-	CreateObjects int
-	RandomRanges  bool
-	Collector     *Collector
-	objects       generator.Objects
-	Versions      int
+	CreateObjects     int
+	RandomRanges      bool
+	Collector         *Collector
+	objects           generator.Objects
+	Versions          int
```
klauspost, line 40: Versions are ignored when using existing objects. 

You could make it...

A) Use up to "Versions" most recent versions when enabled (my preferred solution) or...
B) Error out when set to a value > 1.

pkg/bench/get.go
```diff
@@ -47,6 +51,56 @@ type Get struct {
...
+			}
+			obj := generator.Object {
+				Name: object.Key,
+				Size: object.Size,
+			}
+			g.objects = append(g.objects, obj)
+
+			// limit to ListingMaxObjects
+			if g.ListingMaxObjects > 0 && len(g.objects) >= g.ListingMaxObjects {
+				break
```
klauspost, line 90: Provide a cancelable ctx to `ListObjects` and cancel it when we return, so it isn't leaking resources while we run the benchmark.

klauspost (commented):
Thanks for the contribution!

Some suggestions:

cli/get.go
```diff
@@ -25,6 +25,24 @@ import (
 )
 
 var getFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name: "get-from-listing-bucket",
```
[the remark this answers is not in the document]
fatpat: fixed

cli/get.go
```diff
@@ -25,6 +25,24 @@ import (
...
+		Usage: "Instead of preparing the bench by PUTing some objects, only use all objects already in the bucket",
+	},
+	cli.IntFlag{
+		Name: "listing-bucket-max-objects",
+		Value: 0,
+		Usage: "When using get-from-listing-bucket, limit the number of object to use for GET (0 means all, this is the default).",
+	},
+	cli.StringFlag{
+		Name: "listing-bucket-prefix",
```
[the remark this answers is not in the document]
fatpat: fixed

cli/get.go
```diff
@@ -25,6 +25,24 @@ import (
...
+		Value: "",
+		Usage: "When using get-from-listing-bucket, list only objects in the set prefix (empty string means no prefix, this is the default).",
+	},
+	cli.BoolFlag{
+		Name: "listing-bucket-no-recursive",
```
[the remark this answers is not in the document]
fatpat: fixed

pkg/bench/get.go
fatpat: fixed

pkg/bench/get.go
fatpat: fixed

cli/get.go
```diff
@@ -25,6 +25,24 @@ import (
 )
 
 var getFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name: "get-from-listing-bucket",
+		Usage: "Instead of preparing the bench by PUTing some objects, only use all objects already in the bucket",
+	},
+	cli.IntFlag{
+		Name: "listing-bucket-max-objects",
```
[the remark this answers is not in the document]
fatpat: fixed

cli/get.go
```diff
@@ -25,6 +25,24 @@ import (
 )
 
 var getFlags = []cli.Flag{
+	cli.BoolFlag{
```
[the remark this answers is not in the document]
fatpat: documented

klauspost:
(go fmt is also needed)

pkg/bench/get.go
fatpat: fixed

pkg/bench/get.go
fatpat: fixed, versioning has been implemented

fatpat:
> 

done :)

klauspost (approved):
LGTM

PR #251, opened 2023-02-28
Leverage Helm values in templates

yaraskm:
- A number of variables exist in values.yaml that are not leveraged at all in the deployment templates, such as:
    - Resources
    - Service Account
    - Affinity
    - Tolerations
    - Security Context

- Updated the templates to leverage values that were already possible, as well as adding support for image.version

### DIFF
--- a/k8s/helm/templates/job.yaml
+++ b/k8s/helm/templates/job.yaml
@@ -28,9 +28,15 @@ spec:
           - name: WARP_REGION
             value: {{ .Values.warpConfiguration.s3ServerRegion | quote }}
           - name: WARP_ACCESS_KEY
-            value: {{ .Values.warpConfiguration.s3AccessKey | quote }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "warp.fullname" . }}-credentials
+                key: access_key
           - name: WARP_SECRET_KEY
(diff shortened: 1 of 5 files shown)


k8s/helm/templates/job.yaml
```diff
@@ -10,11 +10,11 @@ spec:
       restartPolicy: Never
       containers:
       - name: {{ include "warp.fullname" . }}
-        image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
+        image: "{{ .Values.image.repository }}:{{ include "warp.imageVersion" . }}"
```
harshavardhana, line 13: are you trying to propose a breaking change?

harshavardhana (changes requested):

k8s/helm/templates/job.yaml
yaraskm: Sorry for the (heavily) delayed reply. This shouldn't be a breaking change, as the previous logic would only ever default the image tag to `latest`, since that's the hard coded `appVersion` in Chart.yaml. My change here will still default to this behaviour, but if you pass `image.version` in your values file you can now specify a specific version of the Warp image to use with the installation.

PR #260, opened 2023-05-17
remove automatic http2 support, instead opt-in

harshavardhana:


cli/flags.go
```diff
@@ -222,4 +222,9 @@ var ioFlags = []cli.Flag{
 		Usage:  "Disable HTTP Keep-Alive",
 		Hidden: true,
 	},
+	cli.BoolFlag{
+		Name:   "enable-http2",
```
klauspost, line 226:
```suggestion
		Name:   "http2",
```

cli/print.go
```diff
@@ -55,7 +55,7 @@ func printInfo(data ...interface{}) {
 	} else {
 		data = append(data, "\n")
 	}
-	console.Info(data...)
+	console.Infoln(data...)
```
klauspost, line 58: What is this solving? I am pretty sure it will break output. We should already add a `\n` above if we aren't printing to console.

klauspost (changes requested):

cli/print.go
klauspost: The line above adds a newline on non-console.

For console output, we want to overwrite the line to keep output clean.

harshavardhana: oh weird but why does `Cleanup Done.` never adds a new line @klauspost ?

klauspost (approved):

PR #285, opened 2023-09-13
check bucket locking status before clean up

### LINKED ISSUE #284: clean up: always sending x-amz-bypass-governance-retention
Hi and thank you so much for this very useful tool, it is really awesome!

It seems there's an issue with the cleanup, it always sends the `x-amz-bypass-governance-retention` header, but AWS or other tools will respond with an exception:`Invalid argument: x-amz-bypass-governance-retention is only applicable to Object Lock enabled buckets.`

It seems it is here:
https://github.com/minio/warp/blob/3dd01804566ab8dbfab35b1fe38e59b7ff02f1eb/pkg/bench/benchmark.go#L218

One fix would be to validate if the bucket has it enabled beforehand with a call to `GeObjectLockConfiguration`. I'm not very proficient in go, I could try to have a look. 

Thanks again for this excellent tool!

bentsku:
This would fix #284, this checks the bucket `ObjectLockConfiguration`,  and if one is present, sets the proper `RemoveObjectsOptions` field `GovernanceBypass`, as it can't normally be passed if the bucket is not "locked". 

This is my first time with Go, so sorry in advance if the code is not following best practices or is not idiomatic.

### DIFF
--- a/pkg/bench/benchmark.go
+++ b/pkg/bench/benchmark.go
@@ -215,7 +215,13 @@ func (c *Common) deleteAllInBucket(ctx context.Context, prefixes ...string) {
 		}
 	}()
 
-	errCh := cl.RemoveObjects(ctx, c.Bucket, objectsCh, minio.RemoveObjectsOptions{GovernanceBypass: true})
+	delOpts := minio.RemoveObjectsOptions{}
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #290, opened 2023-12-06
[delete] bench object from listing the bucket

fatpat:
backport listing feature from get to delete

when doing a delete bench, during the prepare phase, instead of putting objects in the bucket use the objects already in there. This can be used to bench a specific use case with an already filled in bucket.

New options:
--list-existing to activate the listing bucket (current
  behaviour remains)
--list-flat: disable recursive listing


pkg/bench/delete.go
```diff
@@ -37,89 +37,139 @@ type Delete struct {
...
-					mu.Lock()
-					if groupErr == nil {
-						groupErr = err
+		// ensure the bucket exist
+		found, err := cl.BucketExists(ctx, d.Bucket)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return (fmt.Errorf("bucket %s does not exist and --list-existing has been set", d.Bucket))
```
klauspost, line 60:
```suggestion
			return fmt.Errorf("bucket %s does not exist and --list-existing has been set", d.Bucket)
```

pkg/bench/delete.go
```diff
@@ -37,89 +37,139 @@ type Delete struct {
...
+				break
+			}
+		}
+		if len(d.objects) == 0 {
+			return (fmt.Errorf("no objects found for bucket %s", d.Bucket))
+		}
+		done()
+		d.Collector = NewCollector()
+
+	} else { // prepare the bench by creating the bucket and pushing some objects
```
klauspost, line 93: Just shuffle and return. Remove `else`.

klauspost (commented):
Just  some minor things.

If/when you want to do another one, we should see if we can generalize the code. I feel like almost everything in `if d.ListExisting {` could be generalized.

pkg/bench/delete.go
fatpat: fixed

pkg/bench/delete.go
fatpat: fixed

fatpat:
> If/when you want to do another one, we should see if we can generalize the code. I feel like almost everything

I agree, were should I put the factorized code ?

klauspost (approved):
LGTM

harshavardhana (approved):

PR #292, opened 2023-12-19
update dependencies and introduce --disable-sha256-payload

harshavardhana:


cli/multipart.go
```diff
@@ -101,6 +101,7 @@ func multipartOpts(ctx *cli.Context) minio.PutObjectOptions {
 	return minio.PutObjectOptions{
 		ServerSideEncryption: newSSE(ctx),
 		DisableMultipart:     false,
+		DisableContentSha256: ctx.Bool("disable-sha256-payload"),
```
vadmeste, line 104: Since we have cli.Context here; maybe we can calculate if tls is being used or not and disable sha256 automatically if tls is enabled

harshavardhana: That is done via minio-go @vadmeste this is when non-TLS is being used.

vadmeste (approved):
LGTM

PR #294, opened 2024-01-12
check the bucket is empty before clean the bucket

rluisr:
I had run warp on my production bucket and deleted all my customer data.

I should have run the help command and used `noclear` flag or created a `warp-benchmark-bucket` bucket.

Let me take a proposal, If you specify a bucket and let the user suggest a feature to continue or not if the bucket is not empty.

```
./warp put ...
warp: Preparing server.                                                                                                                                                                                                                                                                                                                                                               
bucket: <masked> is not empty. Do you want to continue with cleaning bucket? (y/n)
```

### DIFF
--- a/pkg/bench/benchmark.go
+++ b/pkg/bench/benchmark.go
@@ -114,9 +114,49 @@ func (c *Common) ErrorF(format string, data ...interface{}) {
 	c.Error(fmt.Sprintf(format, data...))
 }
 
+// checkBukectIsEmpty will check if a bucket is empty.
+// If not empty it will ask the user if it should be cleared.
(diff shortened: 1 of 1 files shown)


harshavardhana:
Sir, you are a classic example of not reading `--help` 

```
../warp/warp put -h | grep bucket
  --bucket value         Bucket to use for benchmark data. ALL DATA WILL BE DELETED IN BUCKET! (default: "warp-benchmark-bucket")
  --noclear              Do not clear bucket before or after running benchmarks. Use when running multiple clients.
```

I do not think we need any interactive behavior here.

PR #302, opened 2024-03-06
get: Add --range-size to have a random but a defined range pattern

vadmeste:
--range-size and --range will create GET requests with random offset but with fixed range

### DIFF
--- a/cli/get.go
+++ b/cli/get.go
@@ -53,6 +53,10 @@ var getFlags = []cli.Flag{
 		Name:  "range",
 		Usage: "Do ranged get operations. Will request with random offset and length.",
 	},
+	cli.StringFlag{
+		Name:  "range-size",
+		Usage: "Use a fixed range size while doing random range offsets, --range is implied",
+	},
 	cli.IntFlag{
 		Name:  "versions",
 		Value: 1,
@@ -81,11 +85,22 @@ FLAGS:
 // mainGet is the entry point for get command.
 func mainGet(ctx *cli.Context) error {
 	checkGetSyntax(ctx)
+
+	var rangeSize int64
+	if rs := ctx.String("range-size"); rs != "" {
+		s, err := toSize(rs)
+		if err != nil {
+			return err
+		}
+		rangeSize = int64(s)
+	}
+
 	sse := newSSE(ctx)
 	b := bench.Get{
 		Common:        getCommon(ctx, newGenSource(ctx, "obj.size")),
 		Versions:      ctx.Int("versions"),
-		RandomRanges:  ctx.Bool("range"),
+		RandomRanges:  ctx.Bool("range") || ctx.IsSet("range-size"),
+		RangeSize:     rangeSize,
 		CreateObjects: ctx.Int("objects"),
 		GetOpts:       minio.GetObjectOptions{ServerSideEncryption: sse},
 		ListExisting:  ctx.Bool("list-existing"),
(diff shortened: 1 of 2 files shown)


pkg/bench/get.go
```diff
@@ -292,10 +293,16 @@ func (g *Get) Start(ctx context.Context, wait chan struct{}) (Operations, error)
...
-					start := rng.Int63n(op.Size - size)
-					end := start + size
+					var start, end int64
+					if g.RangeSize == 0 {
+						// Randomize length similar to --obj.randsize
+						size := generator.GetExpRandSize(rng, 0, op.Size-2)
+						start = rng.Int63n(op.Size - size)
+						end = start + size
+					} else {
+						start = g.RangeSize * rng.Int63n(op.Size/g.RangeSize)
```
klauspost, line 303: Why not a random start?
```suggestion
						start = rng.Int63n(op.Size - g.RangeSize)
```

pkg/bench/get.go
```diff
@@ -292,10 +293,16 @@ func (g *Get) Start(ctx context.Context, wait chan struct{}) (Operations, error)
 				}
 
 				if g.RandomRanges && op.Size > 2 {
-					// Randomize length similar to --obj.randsize
-					size := generator.GetExpRandSize(rng, 0, op.Size-2)
-					start := rng.Int63n(op.Size - size)
-					end := start + size
+					var start, end int64
+					if g.RangeSize == 0 {
```
klauspost, line 297:
```suggestion
					if g.RangeSize <= 0 {
```

vadmeste:
We can discuss during the next week if we should drop this PR or not @klauspost

klauspost:
Lint fix at: https://github.com/minio/warp/pull/303

cli/get.go
```diff
@@ -43,10 +43,18 @@ var getFlags = []cli.Flag{
 		Value: "10MiB",
 		Usage: "Size of each generated object. Can be a number or 10KiB/MiB/GiB. All sizes are base 2 binary.",
 	},
+	cli.StringFlag{
+		Name:  "obj.name",
+		Usage: "Specify an object name",
+	},
```
klauspost, line 49: In terms of keeping this simple - could you explain why this is needed?

It seems like it would break pretty much all scenarios since they upload multiple objects.

vadmeste: This is originally written to benchmark server caching, so warp clients will read the same object multiple times, but yeah I think it is better if we make the orchestor warp client to generate a list of objects and share it with other warp clients before starting to benchmark.

klauspost: The multipart benchmark does this already: https://github.com/minio/warp/blob/1723cdbb9c71c2db26e14ed905a89da8bbd29fa0/cli/multipart.go#L94

klauspost: You may want the initiator to create the objects - otherwise they will all be fighting to create it.

harshavardhana:
is this PR still planned? if not please close.

klauspost:
What is the point of the single object name?

You can use `--objects=1` (which will create one per client) or `--list-existing`... which will just use whatever is there.

harshavardhana:
The requirement is that have a bunch of large objects for example and then you can specify a range warp decides to generate list of range requests that randomly hit these objects and provide results, that's all.

klauspost:
@harshavardhana Yes. But not sure why the custom name would be needed for that (and it will break in distributed). To me it seems like it should just be removed from this PR.

harshavardhana (approved):

PR #311, opened 2024-04-05
analysis: add space before printing filename

benpbrown:
Currently, the output looks like:

```
Aggregated data saved toFILENAME.txt
```

There should be a space after 'to', so that's what this PR adds.

### DIFF
--- a/cli/analyze.go
+++ b/cli/analyze.go
@@ -244,7 +244,7 @@ func printAnalysis(ctx *cli.Context, o bench.Operations) {
 		} else {
 			f, err := os.Create(fn)
 			fatalIf(probe.NewError(err), "Unable to create create analysis output")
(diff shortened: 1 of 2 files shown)


klauspost:
Println should already add spaces.

See output from `λ warp analyze -analyze.out=out.csv "warp-get-2023-01-12[133014]-nbUJ.csv.zst"|more` - where the space is added.

Seems to be an [issue](https://github.com/fatih/color/issues/218) with the terminal color printer.

klauspost:
I've proposed an upstream fix here: https://github.com/fatih/color/pull/228

klauspost (changes requested):
You can change the `go.mod` to point to the [commit with the fix](https://github.com/fatih/color/commit/b6598b12a645b3159c1bce51b0e3fafc269510be).

PR #334, opened 2024-08-27
Update goreleaser to support more architectures

allanrogerr:
This PR supports the addition of more architectures to the Warp docker repository.

It keeps the original docker.com releases of `amd64`, while adding new `amd64` and `arm64` images at `quay.io`.


pjuarezd:
On this PR new warp images will be pushed to `quay.io` registry, (previously was only docker hub registry), for your conscideration @donatello

pjuarezd (approved):

harshavardhana (approved):

PR #340, opened 2024-10-04
Fix hang at end of distributed warp with influxdb

jasoncwik:
When running distributed warp with `--influxdb`, the app would always hang at the end of each run. After investigation, it appears this is because the channel used to send operations to Influx isn't used on the master process and thus, never gets closed. The app then hangs on the global wait group waiting for the influx goroutine to complete.

As a fix, I closed the `ExtraOut` channels after the `runServerBenchmark` completes. Alternatively, you could not start the influx channel on the master node.


klauspost (approved):
LGTM. Thanks for the contribution!

PR #343, opened 2024-10-29
add support for stat with --list-existing

harshavardhana:


cli/stat.go
```diff
@@ -72,6 +80,12 @@ func mainStat(ctx *cli.Context) error {
 		StatOpts: minio.StatObjectOptions{
 			ServerSideEncryption: sse,
 		},
+		ListExisting: ctx.Bool("list-existing"),
+		ListFlat:     ctx.Bool("list-flat"),
+		ListPrefix:   ctx.String("prefix"),
+	}
+	if b.ListExisting {
+		b.CreateObjects = 0
```
klauspost, line 88: "CreateObjects" acts as a limit in other `list-existing` implementations. This should be similar to that.

pkg/bench/stat.go
```diff
@@ -35,15 +35,87 @@ type Stat struct {
...
+				}
+
+				if version, found := versions[object.Key]; found {
+					if version >= g.Versions {
+						continue
+					}
+					versions[object.Key]++
+				} else {
+					versions[object.Key] = 1
+				}
```
klauspost, line 101:
```suggestion
				}
				versions[object.Key]++
				
```

pkg/bench/stat.go
```diff
@@ -35,15 +35,87 @@ type Stat struct {
...
+				obj.VersionID = object.VersionID
+			}
+
+			g.objects = append(g.objects, obj)
+
+			// limit to ListingMaxObjects
+			if g.CreateObjects > 0 && len(g.objects) >= g.CreateObjects {
```
klauspost, line 108: Used here...

cli/get.go
```diff
@@ -107,9 +107,6 @@ func mainGet(ctx *cli.Context) error {
 		ListFlat:      ctx.Bool("list-flat"),
 		ListPrefix:    ctx.String("prefix"),
 	}
-	if b.ListExisting && !ctx.IsSet("objects") {
-		b.CreateObjects = 0
-	}
```
harshavardhana, line 112: this is not needed since `--objects` if specified does the actual limit, we do not have to zero out CreateObjects() when `--objects` is not set it is zero anyway.

klauspost: ok

klauspost (approved):

PR #344, opened 2024-10-29
add waitgroup and write the csv.zst in parallel

harshavardhana:


klauspost (changes requested):
This is not safe, as ops will be mutated while analyzing.

PR #348, opened 2024-11-19
dont check --batch/--concurrent/--objects consistency with --list-existing

fatpat:
when `--list-existing` is set there's not need to ensure `--batch`, `--concurrent` and `--objects` values are consistent.


klauspost (approved):

fatpat:
I've added an exclusion on the linter and #350 cames up in the mean time. I'll remove the last commit once the PR is mergerd

PR #350, opened 2024-11-19
fix: min,max keywords across codebase and move to go1.23

harshavardhana:


klauspost:
Hmmm... Thought I already did this one... Anyway, seems like one slipped...

harshavardhana:
PTAL @klauspost

PR #349, opened 2024-11-19
add --metadata and --tag during put operations

fatpat:
- the new `--metadata key=value` flag allow to set user metadata during write operations.
- the new `--tag key=value` flag allow to set user tag during write operations.

this can be used to test some S3 services with more complex scenarios

PR #354, opened 2024-12-23
fix: warp concurrent crash on PUT

harshavardhana:
A regression caused by #349

### DIFF
--- a/pkg/bench/put.go
+++ b/pkg/bench/put.go
@@ -70,7 +70,18 @@ func (u *Put) Start(ctx context.Context, wait chan struct{}) (Operations, error)
 		go func(i int) {
 			rcv := c.Receiver()
 			defer wg.Done()
+
+			// Copy usermetadata and usertags per concurrent thread.
 			opts := u.PutOpts
+			opts.UserMetadata = make(map[string]string, len(u.PutOpts.UserMetadata))
+			opts.UserTags = make(map[string]string, len(u.PutOpts.UserTags))
+			for k, v := range u.PutOpts.UserMetadata {
+				opts.UserMetadata[k] = v
+			}
+			for k, v := range u.PutOpts.UserTags {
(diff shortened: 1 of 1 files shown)


pkg/bench/put.go
```diff
@@ -70,7 +70,18 @@ func (u *Put) Start(ctx context.Context, wait chan struct{}) (Operations, error)
...
+			// Copy usermetadata and usertags per concurrent thread.
 			opts := u.PutOpts
+			opts.UserMetadata = make(map[string]string, len(u.PutOpts.UserMetadata))
+			opts.UserTags = make(map[string]string, len(u.PutOpts.UserTags))
+			for k, v := range u.PutOpts.UserMetadata {
+				opts.UserMetadata[k] = v
+			}
+			for k, v := range u.PutOpts.UserTags {
+				opts.UserTags[k] = v
+			}
```
ramondeklein, line 83: Not sure why this was causing a problem. I would guess that `UserMetadata` and `UserTags` are read-only field and are not written. The shared map should be fine. 

If the maps are altered from outside `Put.Start`, then it may be safer to perform this copy operation outside the go-routine.

harshavardhana: It must be copied per goroutine before Put.

harshavardhana: Which is what I am doing.

harshavardhana: There is a write happening because the user metadata is used perhaps in MinIO-go to add other metadata (default) per request it can't be shared.

ramondeklein: I thought that the caller of this method would change the map. In that case it should be put outside the `go` routine on line 70. But if it's changed later, then this is fine.

ramondeklein (approved):

klauspost:
It mutates them?  oh, wow.

PR #362, opened 2025-01-30
Make data collection aggregated throughout

klauspost:
This PR will change all data collection to be automatically aggregated as requests happen.

This means individual requests will no longer be recorded unless `-full` is specified.

This will allow warp to run for days without consuming significant amount of memory. The only downside is that there are no full-run median request values, but they are instead grouped and averaged.

* Replaces output display with nicer one.
* Displays live aggregated values while running.
* Allows stopping running benchmarks and save result.
* Displays remote data live.
* Allows `--autoterm` for remote running benchmarks as well.
* `-analyze.v` does not do anything (all details shown by default)
* Adds (hidden) `--stdout` to print operations to stdout.
* New output format is a zstd compressed JSON file.

Obviously a significant rewrite, so there will likely be uncovered corner cases. But testing checks out fine now.

Pending:

* [ ] Update documentation.
* [ ] Update YAML files.
* [ ] Update `warp cmp`. Not strictly needed for release.
* [ ] Update `warp merge`. Rarely used.


klauspost:
@harshavardhana If you port it to Windows ;)

Let's see if this works: https://github.com/user-attachments/assets/c68b74d1-1748-4e1b-89f4-bbb455c4baff

harshavardhana:
> @harshavardhana If you port it to Windows ;)
> 
> Let's see if this works: https://github.com/user-attachments/assets/c68b74d1-1748-4e1b-89f4-bbb455c4baff

why do you have a more verbose output for default results when the test is over? can we have the same output as it is total which mainly talks about the higher level points?

klauspost:
@harshavardhana Brought back the simple view - thinking of it - it was only on `warp analyze` it annoyed me, so I kept the "always full detail" there.

harshavardhana (approved):

PR #364, opened 2025-02-04
Fix inconsistent output

klauspost:
* Fixes fluctuating request times during benchmark.
* Fix wrong min/max time on requests in report.
* Hide "current" if data is older than 15s.
* Don't show 0-byte operation throughput.
* Make byte counter neater.

Fixes:

* Race in mixed benchmark.
* Crash when operation with early start arrives.

### DIFF
--- a/cli/ui.go
+++ b/cli/ui.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/pprof"
+	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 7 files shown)


harshavardhana (approved):

harshavardhana (approved):

klauspost:
(let's hold this of.. got a bit more)

klauspost:
That should be it for now.

PR #377, opened 2025-02-26
Update the go version in Dockerfile

chenhuazhong:
golang:1.22 --> golang:1.23


chenhuazhong:
@klauspost

chenhuazhong:
@harshavardhana

klauspost:
@chenhuazhong Please stop bugging people. We will get to it when we get to it.

klauspost:
We use the toolchain now, so it shouldn't really matter which version is used.

PR #383, opened 2025-03-25
Support multipart-put benchmarking

d1tyo:
Proposed solution (`warp multipart-put`) is aimed to benchmark uploading of multiple multipart objects concurrently. It periodically creates new multipart uploads and concurrently uploads parts.


harshavardhana (changes requested):
You don't like `warp multipart` ?

d1tyo:
> You don't like `warp multipart` ?

`warp multipart` doesn't allow to benchmark multipart **upload** itself. It mostly benchmarks download and only uploads to a single multipart upload id.

Proposed solution is aimed to benchmark uploading of multiple multipart objects concurrently. It periodically creates new multipart uploads and concurrently uploads parts.

So proposed `warp multipart-upload` benchmarks a different scenario from `warp multipart`.

cli/multipart_upload.go
```diff
@@ -0,0 +1,71 @@
...
+		Value: 20,
+		Usage: "Run this many concurrent operations per each multipart upload. Must not exceed obj.size/part.size",
+	},
+}
+
+var MultiPartUploadCombinedFlags = combineFlags(globalFlags, ioFlags, multipartUploadFlags, genFlags, benchFlags, analyzeFlags)
+
+// MultipartUpload command
+var multipartUploadCmd = cli.Command{
+	Name:   "multipart-upload",
```
klauspost, line 31: Let's be a bit consistent in our naming.
```suggestion
	Name:   "multipart-put",
```

pkg/bench/multipart_upload.go
```diff
@@ -0,0 +1,178 @@
...
+	PartsNumber      int
+	PartsConcurrency int
+}
+
+// Prepare for the benchmark run
+func (g *MultipartUpload) Prepare(ctx context.Context) error {
+	if err := g.createEmptyBucket(ctx); err != nil {
+		return err
+	}
+	return nil
```
klauspost, line 27:
```suggestion
	return g.createEmptyBucket(ctx)
```

pkg/bench/multipart_upload.go
```diff
@@ -0,0 +1,178 @@
...
+					return err
+				}
+
+				obj := g.Source().Object()
+				client, done := g.Client()
+				defer done()
+				core := minio.Core{Client: client}
+				op := Operation{
+					OpType:   http.MethodPut,
+					Thread:   thread,
```
klauspost, line 136: I honestly do not know how the stats will behave with the same thread id returning overlapping entries.

I do see the value of individually reporting uploads, but I am not sure it is safe with the outputted stats.

A "safer" approach would be to report the per object time - which would then of course also include Create/Complete operations.

But if you test, do what makes sense to you.

pkg/bench/multipart_upload.go
```diff
@@ -0,0 +1,178 @@
...
+		thread := uint16(i)
+		eg.Go(func() error {
+			<-wait
+
+			for {
+				select {
+				case <-ctx.Done():
+					return nil
+				default:
+				}
```
klauspost, line 49: As in your other code...

```suggestion
			for ctx.Err() == nil {
```

(you would need a return at the end, though)

cli/multipart_upload.go
```diff
@@ -0,0 +1,71 @@
...
+	if ctx.NArg() > 0 {
+		console.Fatal("Command takes no arguments")
+	}
+	if ctx.Bool("disable-multipart") {
+		console.Fatal("cannot disable multipart for multipart-upload test")
+	}
+
+	if ctx.Int("parts") > 10000 {
+		console.Fatal("parts can't be more than 10000")
+	}
```
klauspost, line 70:
```suggestion
	if ctx.Int("parts") > 10000 {
		console.Fatal("parts can't be more than 10000")
	}
	if ctx.Int("parts") <= 0 {
		console.Fatal("parts must be at least 1")
	}
```

klauspost: Check "part.size" and "part.concurrent"

pkg/bench/multipart_upload.go
```diff
@@ -0,0 +1,178 @@
...
+					op.File = ""
+				}
+
+				opts := minio.PutObjectPartOptions{
+					SSE:                  g.Common.PutOpts.ServerSideEncryption,
+					DisableContentSha256: g.PutOpts.DisableContentSha256,
+				}
+
+				op.Start = time.Now()
+				_, err := core.PutObjectPart(nonTerm, g.Bucket, objectName, uploadID, partIdx, obj.Reader, obj.Size, opts)
```
klauspost, line 152: Sanity check size.

klauspost (commented):
Sorry about the name change. Rest should be trivial.

cli/multipart_upload.go
d1tyo: Done. Please note that unlike an old `multipart` mode I do not add a minimal of 5MiB constraint to `part.size` because there are some S3 implementations that allow it to be less than 5MiB.

pkg/bench/multipart_upload.go
d1tyo: Could you please clarify what you mean?

pkg/bench/multipart_upload.go
d1tyo: Done.

pkg/bench/multipart_upload.go
d1tyo: Thanks for noticing. Actually it's hard to tell if it works properly. Numbers in stats look OK, but are they really? Idk.

Now I changed it to `thread*g.PartsConcurrency + i`, so each thread will have its own number and they must not collide (supposing `PartsConcurrency*Concurrency < 65536`). I believe the formula is clear, but in case it is not: `g.PartsConcurrency` here is like a stride for 2d matrix, `thread` - is Y-coordinate and `i` is X-coordinate

d1tyo:
Just updated PR. I renamed both multipart_upload.go files to multipart_put.go as well.
Please take a look.

pkg/bench/multipart_upload.go
klauspost: The call returns the object size. Check that it matches the size you expect. For example:

https://github.com/minio/warp/blob/483f844249cb53dd6be68f688c3aa4b2869abd0f/pkg/bench/put.go#L131-L138

pkg/bench/multipart_upload.go
klauspost: To some extent I think the most genuine is to do full object timings. I don't particularly like that it shows the wrong objects/sec.

But I see what you are trying to measure.

Let's name the operation `PUTPART` instead of just PUT, so it is clear that we are looking at something else. If shouldn't really be a bigger change than changing it in the operation.

cli/multipart_upload.go
klauspost: Not a massive fan, but ok.

pkg/bench/multipart_upload.go
d1tyo: Oh yes, that makes sense. Added the check.

pkg/bench/multipart_upload.go
d1tyo: Alright, I renamed it to `PUTPART` inplace. I think it is better to make it a constant in pkg/bench/ops.go. Like that:
```go
type opType string

const (
  opTypePutPart opType = "PUTPART"
)
```

But in that case it is probably better to do so everywhere and change `Operation.OpType` type to `opType`. Although probably it's better to do in a separate PR. What do you think? I can do it in a follow-up pr; it doesn't seem like it takes too much effort.

d1tyo:
Updated it once again.

klauspost (approved):
ltgm 👍🏼

pkg/bench/multipart_upload.go
klauspost: It is just a string, with no restrictions other than being sensible to the operation.

harshavardhana (approved):

klauspost:
2 minor lint issues. We can merge if those get resolved.

d1tyo:
Just resolved it.

harshavardhana (approved):

PR #399, opened 2025-05-17
support warp to use S3Express, EC2 instance credentials

harshavardhana:


klauspost (approved):
LGTM. You should probably add it to the README.

harshavardhana:
> LGTM. You should probably add it to the README.

Will do have to rewrite it a bit.

PR #401, opened 2025-06-04
Experimental: Add Kernel TLS

klauspost:
Attempt to use Kernel TLS if available and using `-ktls` parameter.

* Enable Kernel TLS: `sudo modprobe tls`
* Disable KTLS: `sudo modprobe -r tls`


klauspost:
Definitely not comfortable adding it server-side.

Linux users, check it out.

zveinn (approved):

PR #406, opened 2025-06-30
Improve partial compression

widavis:
Previously the Partial Compression option would use all 0s for the compressible regions. This is undesirable as the only mechanism; now we default to using randomized text strings, though "all 0s" is also an option


widavis:
Opened by mistake

PR #422, opened 2025-10-07
fix: fatal errors break the terminal due to incorrect bubble tea shutdown

marktheunissen:
Make the correct ui.Wait() call when exiting on a fatal log, to prevent the terminal cursor from disappearing, as well as other problems that make the terminal unusable:

Before:

Screenshot 2025-10-07 at 15 49 35

After:

Screenshot 2025-10-07 at 15 49 59

### DIFF
--- a/cli/print.go
+++ b/cli/print.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"unicode"
 
 	"github.com/cheggaaa/pb"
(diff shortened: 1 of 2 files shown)


cli/print.go
```diff
@@ -63,6 +71,10 @@ func fatalIf(err *probe.Error, msg string, data ...any) {
 	if err == nil {
 		return
 	}
+	// Clean up UI if one is registered
+	if u := globalUI.Load(); u != nil {
+		u.Wait()
+	}
```
marktheunissen, line 77: It's not ideal, but the problem is fatalIf does exit(). Another solution would be to refactor fatalIf itself, but it's used everywhere...

harshavardhana (approved):

klauspost (approved):

PR #423, opened 2025-10-09
fix put --add-metada

fatpat:
--add-metadata is not working


klauspost (approved):

zhsensen:
It's great to support adding metadata，Can the md5 value of an object be added to metadata?

klauspost:
@zhsensen Just read the help. `--md5`.

PR #425, opened 2025-10-31
Update go and dependencies

bazko1:
Update go 1.24 -> 1.25
Update subpackages including small breaking change in `golang.org/x/crypto`
Fix non format print clauses that used formatting and new version of linter caught.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,18 +31,6 @@ linters:
   exclusions:
     generated: lax
     rules:
-      - linters:
-          - govet
-        text: "Println call has possible Printf formatting"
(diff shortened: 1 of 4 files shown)


klauspost:
Seems like you won the task of fixing new linter issues :)

klauspost:
@bazko1 That is not "fixing". At least one was quite genuine.

bazko1:
> @bazko1 That is not "fixing". At least one was quite genuine.

Ups my bad I assumed console support formatting in all functions and new linter started to be wrong :eyes:

klauspost (approved):
There we go :D

PR #426, opened 2025-11-06
Highlight the danger of data being cleaned

KhajiitJzargo:
I made a big mistake locally, that I suspect others might fall victim to. As such, I propose to highlight the danger of data being wiped with the warning-format. If that is to bold, then I suggest the note variant. Either way, I would argue something needs to be done to warn users of potential data loss, and not in the form of a last sentence in an arbitrary paragraph with an italicized "not".

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ By default operations are performed on a bucket called `warp-benchmark-bucket`.
 This can be changed using the `--bucket` parameter. 
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


klauspost (commented):
Not sure it will make any difference. Seems like copium ;)

Anyway, I don't see any problem with it.  Let's make it a bit clearer as a standalone sentence.

PR #446, opened 2026-01-06
Support STS web-token based access.

ramondeklein:
## Summary
Support authenticating using an web token (using STS).

## Context
Running Warp should be possible from AIStor operator (using `Warp` job), but the operator is created in such a way that it doesn't require root credentials anymore. To enable running `Warp` using STS is most convenient.

An alternative solution is to fetch the STS issued credentials from the operator and pass it using access-key, secret-key and session-token. However, that is a bit more complicated (due to a stateless reconcile loop). Supporting STS from Warp is simpler and can't do much harm.

## How to test
Invoke Warp with `--signature sts_web_token` and either `--sts-web-token`.

### DIFF
--- a/pkg/aggregate/live.go
+++ b/pkg/aggregate/live.go
@@ -606,6 +606,14 @@ func Live(ops <-chan bench.Operation, updates chan UpdateReq, clientID string, e
 	}
 	lastUpdate := time.Now()
 	for op := range ops {
+		dur := op.End.Sub(op.Start)
+		if dur < 0 {
+			op.End = op.Start
+			op.Err += "Negative duration"
+		} else {
+			// Rewrite op.End in case a non-monotonic adjustment has been made.
+			op.End = op.Start.Add(dur)
+		}
 		if reset.CompareAndSwap(true, false) {
 			a = newRealTime()
 		}


cli/client.go
```diff
@@ -163,8 +163,38 @@ func getClient(ctx *cli.Context, host string) (*minio.Client, error) {
...
 		})
+	case "STS_WEB_TOKEN":
+		var err error
+		stsEndPoint := ctx.String("sts-endpoint")
+		if stsEndPoint == "" {
+			proto := "http"
+			if ctx.Bool("tls") || ctx.Bool("ktls") {
+				proto = "https"
+			}
+			stsEndPoint = fmt.Sprintf("%s://%s", proto, host)
```
klauspost: Is it normal that the sts endpoint tls is tied to the server tls settings? Seems they can/should be separate?

Not an expert here, but it seems strange.

cli/flags.go
```diff
@@ -186,9 +186,24 @@ var ioFlags = []cli.Flag{
...
+	cli.StringFlag{
+		Name:   "sts-web-token",
+		Usage:  "Specify the STS web token",
+		EnvVar: appNameUC + "_STS_WEB_TOKEN",
+	},
+	cli.StringFlag{
+		Name:   "sts-web-token-file",
+		Usage:  "Specify path to a file containing the STS web token",
+		EnvVar: appNameUC + "_STS_WEB_TOKEN_FILE",
+	},
```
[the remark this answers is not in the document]
ramondeklein: Good one. I'll change that.

cli/client.go
ramondeklein: Okay, I'll remove it. If we ever need it, then we can add it.

PR #481, opened 2026-04-09
Update pkg

klauspost:
Fixes #471


rraulinio (approved):

ramondeklein (approved):
Fine if we agree that we can use `minio/pkg/v3` based on commit instead of tag.

PR #486, opened 2026-05-04
Fix RPS limit

klauspost:
RPS limit would ignore concurrency. Allow bursts up to the concurrency, so if RPS limit isn't hit it will operate as normal.

* Add RPS limit to fanout.

* Remove RPS from versioned/mixed since operations aren't part of the recorded benchmark.

Verified other RPS limits.

Replaces https://github.com/minio/warp/pull/485

### DIFF
--- a/cli/benchmark.go
+++ b/cli/benchmark.go
@@ -681,6 +681,9 @@ func checkBenchmark(ctx *cli.Context) {
 			fatalIf(errDummy(), "autoterm.pct cannot be zero or negative")
 		}
 	}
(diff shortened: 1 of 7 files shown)


harshavardhana (approved):

crjvoyager:
@klauspost I downloaded your PR and compiled and tested and it works.  It held all of the connections open and I could see them on my haproxy.

crjvoyager (approved):
I downloaded your PR and compiled and tested and it works. It held all of the connections open and I could see them on my haproxy.

PR #494, opened 2026-07-02
[Duplicate] Add --dns-cache-ttl to avoid DNS storm at high concurrency

kojiwell:
With CGO_ENABLED=0 (warp's release build), Go uses the pure resolver which does no in-process DNS caching and performs a fresh lookup on every dial. At high --concurrent against an FQDN host this produces a burst of DNS queries per connection ("DNS storm").

Add an opt-in --dns-cache-ttl flag backed by a process-wide TTL cache: the host FQDN is resolved at most once per TTL regardless of concurrency or the number of clients created by host selection. The FQDN is kept for TLS SNI, certificate validation and the Host header; only the TCP dial address is substituted with a cached IP.

Details:
- Lazy re-resolution with single-flight on TTL expiry; on resolution failure the last successful set is kept so the benchmark is not interrupted.
- Multiple A/AAAA records are cached and handed out round-robin per dial.
- Literal IP hosts are a no-op; ttl 0 (default) disables caching and preserves the previous behavior exactly.
- Applied to all transports (default / TLS / kTLS-h2 via DialContext, kTLS-non-h2 via a DialTLSContext that pins ServerName=FQDN).
- DurationFlag propagates automatically to distributed clients.

## Summary by CodeRabbit

* **New Features**
  * Added a new `--dns-cache-ttl` CLI option to control DNS caching duration.
  * Introduced shared DNS caching for outbound connections to reduce repeated lookups under high concurrency.
  * Improved TLS connection handling so cached IP dialing still preserves the original hostname for secure connections.

* **Bug Fixes**
  * Avoids unnecessary DNS resolution for literal IP addresses.
  * Preserves previously resolved addresses when a refresh fails, helping connections stay stable.

### DIFF
--- a/cli/client_ktls.go
+++ b/cli/client_ktls.go
@@ -18,6 +18,8 @@
 package cli
 
 import (
+	"context"
+	"net"
 	stdHttp "net/http"
 	"os"
 	"time"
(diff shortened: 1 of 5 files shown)


kojiwell:
Sorry, this is a duplicate of #454. Please ignore.

PR #499, opened 2026-07-29
fix(rdma): bind a CUDA context per worker so --rdma=gpu registrations succeed

harshavardhana:
## Description

`--rdma=gpu` reliably crashed with `SIGSEGV` inside `miniocpp_get_object` at anything past trivial concurrency:

```
SIGSEGV: segmentation violation
PC=... sigcode=2 addr=0x746d9f000000        <- exactly the device buffer argument
signal arrived during cgo execution
github.com/minio/minio-go/v7._Cfunc_miniocpp_get_object(..., 0x746d9f000000, 0x1000000, 0x0, 0x0)
```

**Root cause.** `libcufile` resolves the device for `cuFileBufRegister` through the *calling thread's* current CUDA context.
cgo runs a goroutine's calls on whatever OS thread it likes, and a thread that has never issued a CUDA call has no context, so
registration failed intermittently — visible in `cufile.log`, succeeding on one thread and failing on the next:

```
TRACE cufio_rdma:1517 cuFileBufRegister buffer is GPUDirect RDMA capable       [tid A]
ERROR cufio_core:1558 cuFileBufRegister cuCtxGetDevice error:
                      CUDA_ERROR_INVALID_CONTEXT CUdevice: 0                   [tid B]
ERROR cufio_core:1715 cuFile buffer object device memory check failed
```

Registration failing means the transfer drops to minio-cpp's HTTP-into-buffer fallback, which fills the buffer with ordinary
host stores — into a CUDA device pointer. Hence the fault at exactly the buffer address, and hence the concurrency
correlation: concurrency only changes how often a call lands on a context-less thread.

**Fix.** Pin each GET worker to one OS thread and make the primary context current on it before the first transfer.
Also reuse a single sink buffer per worker rather than allocating and freeing a device buffer per operation.

Note the first commit here is a pre-existing local `minio-go` bump that isn't upstream yet; the tested binary was built with
it. Happy to drop it if you'd rather land it separately.

## Motivation and Context

`--rdma=gpu` was unusable above concurrency ~2 — and when it didn't crash it was silently falling back to non-RDMA transfers,
so the numbers it did report were not measuring RDMA.

## How to test this PR?

Against a 4-node AIStor cluster over RoCE, 16MiB objects, single client, with `rdma_dev_addr_list` set to the client NIC:

```bash
warp get --rdma=gpu --host=<node>:9000 --obj.size=16MiB --objects=32 --concurrent=16 --duration=15s
```

| concurrency | before | after |
| ----------- | ------- | ----- |
| 4  | SIGSEGV | 4366 MiB/s |
| 8  | SIGSEGV | 7056 MiB/s |
| 16 | SIGSEGV | 8264 MiB/s |

Server-side `minio_api_rdma_errors_total{category="retry_exceeded"}` stays at 0 across the runs, and `mc admin trace` shows
every GET carrying `x-amz-rdma-token`. Also verified the untagged build (`go build`, no `-tags rdma`) still compiles via the
stub path.

## Summary by CodeRabbit

- **Performance**
  - Improved GET benchmark efficiency by reusing RDMA buffers across requests to reduce per-request allocation overhead.
  - RDMA buffers now automatically grow when larger downloads are encountered.

- **Reliability**
  - Enhanced GPU RDMA stability by ensuring each worker thread has a valid CUDA device/context before RDMA work.
  - RDMA setup/download failures are reported per operation while the benchmark continues.

- **Compatibility**
  - Non-RDMA builds remain fully supported, with GPU/CUDA context handling treated as unavailable.

- **Chores**
  - Updated the Go toolchain version and refreshed dependency versions.


harshavardhana:
Pushed `1745c6d` to clear the failing `Analysis` check, plus a rebase onto master.

**The vulnerabilities were pre-existing on master, not introduced here** — `grpc`, `x/text`, `hamba/avro` and `iceberg-go`
were all at identical versions on `origin/master`. The rebase also dropped a stale commit (master has that minio-go bump
upstream now) which had been regressing `golang.org/x/crypto` v0.52.0 → v0.51.0 on this branch.

```
GO-2026-6061  google.golang.org/grpc   v1.79.3  -> v1.82.1
GO-2026-5970  golang.org/x/text        v0.37.0  -> v0.40.0
GO-2026-5046/5047/5048  hamba/avro/v2  v2.31.0  -> "Fixed in: N/A"
```

The three avro advisories have no fixed release (v2.31.0 is the newest), so they can't be bumped away directly. They come in
transitively via `apache/iceberg-go`, and **v0.6.0 replaced `hamba/avro` with `twmb/avro`**, which clears them.

```
$ govulncheck ./...
No vulnerabilities found.
$ echo $?
0
```

`go vet ./...`, `go test ./...` and the `-tags=rdma` build all pass.

Flagging the scope explicitly: the `iceberg-go` v0.4.1 → v0.6.0 bump is a larger dependency move than a bug-fix PR would
normally carry. It's the only route to a green scan, but I'm happy to split it into its own PR if you'd prefer this one stay
limited to the CUDA context fix.

PR #500, opened 2026-08-11
feat(rdma): ship releasable --rdma builds

harshavardhana:
## Why

Release binaries are built with `CGO_ENABLED=0` and no `rdma` tag, so `--rdma` has never worked in a released warp: `--rdma=gpu` fatals at startup, and `--rdma=cpu` passed validation and then failed on *every operation* with `ErrRDMANotCompiled`.

## What

**Build tags split.** `rdmabuf_gpu.go` unconditionally pulled in `cuda_runtime.h` / `-lcudart`, so `-tags=rdma` required CUDA even for CPU-mode RDMA, which minio-go does not need. Now `-tags=rdma` links libminiocpp and enables `--rdma=cpu`; `-tags=rdma,cuda` adds the CUDA runtime for `--rdma=gpu`. `HasRDMA` / `HasRDMAGPU` gate them independently. This is what makes a distributable CPU-RDMA binary possible.

**Release artifacts.** `scripts/build-rdma.sh` builds minio-cpp with RDMA and packages a self-contained tarball — bundled `lib/` with a `DT_RPATH` of `$ORIGIN/lib`, so no `LD_LIBRARY_PATH` or system install is needed. `release-rdma.yml` publishes `warp-rdma_linux_{amd64,arm64}` and `warp-rdma-gpu_linux_amd64` on a tag; `go-rdma.yml` compiles both tag combinations on every PR, which nothing did before.

**Fixes found on the way:**

- CPU buffers were Go slices handed to cgo and retained across the call. Now `minio.AlignedBuffer`, paired with `FreeAlignedBuffer` — which was missing entirely, so C-heap buffers leaked per operation.
- PUT never bound a CUDA context to its worker thread, so `--rdma=gpu` PUT hit the `CUDA_ERROR_INVALID_CONTEXT` registration failure that 9bc12d9 fixed for GET. PUT now also reuses one grow-only buffer per worker, like GET.
- `--rdma=cpu` was not checked against the build; `--rdma` was accepted by benchmarks that never allocate an RDMA buffer (`mixed` would have reported HTTP numbers as RDMA); ranged GETs silently ignored the range.

## Verification

- Built the real tarball end-to-end against actual minio-cpp. Two link problems surfaced and are fixed in the script: `-rpath-link` for libminiocpp's own cuObj dependencies, and `--disable-new-dtags`, since `DT_RUNPATH` is not inherited by transitive dependencies.
- Unpacked the tarball elsewhere and ran real `put` and `get` benchmarks with `--rdma=cpu` against a throwaway MinIO — ~6.4k PUTs and ~100k GETs through libminiocpp, with `--obj.randsize` to drive the grow-only reuse path. RSS grew 127→169 MB and flattened; no per-op leak.
- Exercised the `--gpu` branch of the script, and confirmed `nvidia-cuda-dev` ships `cuda_runtime.h` and `libcudart.so` on default paths.

## Notes

- RDMA failures fall back to HTTP inside libminiocpp and the operation still succeeds, so warp cannot report whether RDMA was used. `IsRDMAAvailable()` is also optimistic. Both are documented rather than papered over.
- The RDMA tarballs attach to the GitHub release. qreleaser mirrors only the `CGO_ENABLED=0` artifacts to dl.min.io, so the RDMA builds are GitHub-only for now.

## Summary by CodeRabbit

- **New Features**
  - Added S3-over-RDMA support for CPU and GPU-Direct modes.
  - Added reusable RDMA buffers to improve upload efficiency.
  - Added runtime CUDA detection with clear diagnostics when GPU support is unavailable.
  - Added reproducible RDMA release packaging for amd64 and arm64.

- **Bug Fixes**
  - Improved validation for unsupported benchmarks, ranged GETs, and unavailable RDMA connections.

- **Documentation**
  - Added setup, compatibility, build, usage, benchmarking, and release guidance for S3-over-RDMA.


harshavardhana:
## Fixes Applied Successfully

Fixed 5 file(s) based on 18 CodeRabbit feedback item(s): 14 applied, 4 deferred.

**Files modified:**
- `.github/workflows/go-rdma.yml`
- `.github/workflows/release-rdma.yml`
- `scripts/build-rdma.sh`
- `README.md`
- `cli/get.go`

**Commit:** `892e695`

Two items are worth calling out because the fix differs from the suggestion:

- **Bundle the RDMA runtime dependencies** — the underlying claim is right: `libcuobjclient.so.1` needs `libibverbs`/`librdmacm`/`libnuma`/`libmlx5` from the host, so the README's "it finds its libraries on its own" overpromised. But bundling them would be wrong — `libmlx5` is a verbs provider that must match the host's rdma-core and kernel driver. Documented the host prerequisite instead.
- **Normalize archive metadata** — applied. Verified by building twice from identical inputs and confirming byte-identical archives and SHA-256 sums.

Deferred, with reasons:

- **Pin action references to SHAs** — worthwhile, but the repo's other four workflows all use floating tags; pinning only these two would be inconsistent. Better as a repo-wide change.
- **Pass the benchmark context to `PutObject`** — `nonTerm` is a deliberate warp-wide pattern (`get.go` and the non-RDMA PUT do the same) so an in-flight operation completes rather than being cancelled mid-measurement. Not an RDMA regression.
- **Pin minio-cpp** — its `c_api` evolves alongside minio-go, so a hardcoded commit would silently freeze warp against stale native code. `MINIO_CPP_REF` allows a deliberate pin per release. vcpkg is already pinned to `2026.07.29`.
- **Language identifiers on fences** — superseded; the new section's fences are now tagged.

The latest autofix changes are on the `rdma-release` branch.

harshavardhana:
## Fixes Applied Successfully

Fixed 3 file(s) based on 5 CodeRabbit feedback item(s): 4 applied, 1 already obsolete.

**Files modified:**
- `scripts/build-rdma.sh`
- `README.md`
- `.github/workflows/go-rdma.yml`

**Commit:** `dc73b62`

Notes on two of them:

- **Assert that the CUDA-less invocation fails** — a real bug in the smoke test I added. `|| true` discarded the exit status, so the check would have passed had warp ever exited 0 while printing the diagnostic. I confirmed warp exits 1 on a fatal before tightening the assertion, so requiring a non-zero exit is safe rather than a new flake.
- **Run the RDMA race test suite** — added `go test -race -tags=rdma ./...`. I ran it locally against libminiocpp first to be sure the step passes rather than pushing an untested CI change.

Deferred: **Correct the RDMA archive count and GPU extraction path**. That thread was written against the previous commit, when three archives were published. The dlopen change removed the GPU archive entirely — there is now one archive per architecture, the only extraction path is `warp-rdma/warp`, and the text does not describe cuObj as GPU-only.

The latest autofix changes are on the `rdma-release` branch.

PR #501, opened 2026-08-11
build: publish the RDMA build as warp.rdma

harshavardhana:
Follow-up to #500, which shipped the RDMA build itself. This publishes it through the release pipeline as `warp.rdma`, alongside the stock `warp`.

## Using qreleaser's existing RDMA flavor

qreleaser already knows about RDMA builds. Its transform treats `-rdma` in a build id exactly the way it treats `-fips`: it strips the flavor to find the project, expects the binary inside the dist directory to carry the plain name, and suffixes the published files. The container config has carried `rdma_enabled` / `rdma_archs` on the same basis.

So the build id is `warp-rdma` while the binary stays `warp`, and the release directory ends up with:

```
warp                       warp.rdma
warp.v1.5.0                warp.v1.5.0.rdma
+ .sha256sum for each
```

An earlier revision of this PR declared a `variants` entry instead. That was the wrong mechanism for this pipeline — `release-test` built both binaries fine and then failed in the transform with `Binary not found: dist/warp-rdma_linux_amd64_v1/warp`, which is the flavor convention telling us what it expected. Following the existing convention needs nothing from qreleaser that is not already released.

## What this adds

- `.goreleaser/qreleaser.yaml` — the RDMA build: `CGO_ENABLED=1`, `-tags=kqueue,rdma`, linux/amd64.
- `.github/workflows/qreleaser-test.yml` — provisions libminiocpp so `release-test` genuinely exercises the RDMA release path, and asserts the result.
- `.gitignore` — `cufile.log`, which libcufile writes into the working directory whenever an RDMA binary runs.

The build links libminiocpp through cgo, so unlike the stock build it cannot be cross-compiled and is limited to the release host's architecture. CUDA is dlopened rather than linked, so the one binary serves both `--rdma=cpu` and `--rdma=gpu`, and no CUDA package is needed to build it.

## CI now covers the release path

`release-test` previously reached the RDMA build on a bare runner and died on `miniocpp/c_api.h`. It now builds libminiocpp the same way `go-rdma.yml` does — pinned vcpkg, explicit CMake — and then asserts, rather than merely building:

1. `warp.rdma --version` runs from the release layout, so the flavor survived the transform
2. the stock `warp` **refuses** `--rdma=cpu`
3. `warp.rdma` does **not** report "requires the RDMA build"

Both are named `warp` inside their dist directories, so without that third check the flavor could silently be a copy of the stock binary and nothing would notice.

## Build environment

Two review points applied here. goreleaser replaces same-named environment variables rather than merging them — I verified this empirically — so inherited `CGO_CFLAGS`/`CGO_LDFLAGS` are now carried through instead of silently dropped. `MINIOCPP_PREFIX` is also added to the runtime search path so a non-default prefix resolves.

That rpath is deliberately `DT_RUNPATH`, not `DT_RPATH`. `DT_RPATH` outranks `LD_LIBRARY_PATH`, which would stop a user of the published binary pointing at their own libminiocpp. (`DT_RPATH` is right for the self-contained tarball, where the libraries are bundled and cuObj must resolve transitively — that archive is built separately by `scripts/build-rdma.sh`.)

## Verification

Against the released toolkit `RELEASE.2026-08-11T19-37-29Z`: goreleaser produces both builds, its transform yields the layout above with matching versioned names and checksums, both binaries run, only `warp.rdma` accepts `--rdma=cpu`, and `readelf` confirms `RUNPATH` rather than `RPATH`.

## Note for the release host

It needs libminiocpp built with RDMA plus `libibverbs`, `librdmacm` and `libnuma`; set `MINIOCPP_PREFIX` if that is not `/usr/local`. Without them the build fails at the goreleaser stage — before the tag is pushed, so it is recoverable — or set `WARP_SKIP_RDMA=true` to drop it from a release. The published binary resolves libminiocpp and the cuObj libraries from the host at run time.

## Summary by CodeRabbit

* **New Features**
  * Added an optional RDMA-enabled Linux build variant for supported environments.
  * Added RDMA packages for Debian, RPM, and Alpine-based deployments, including required runtime libraries.
  * Added release packaging with version metadata and SHA-256 checksums.

* **Bug Fixes**
  * Prevented generated RDMA build logs from being included in source changes.

* **Tests**
  * Added automated checks confirming standard and RDMA-enabled binaries behave correctly, including validation of RDMA support during network preparation.

### DIFF
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -68,7 +68,7 @@ jobs:
         ../vcpkg/vcpkg install
         cmake . -B ./build \
           -DCMAKE_BUILD_TYPE=Release \
-          -DBUILD_SHARED_LIBS=ON \
+          -DBUILD_SHARED_LIBS=OFF \
           -DCMAKE_INSTALL_PREFIX=/usr/local \
           -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake \
           -DMINIO_CPP_ENABLE_RDMA:BOOL=ON
@@ -76,6 +76,8 @@ jobs:
         sudo cmake --install ./build
         cuobj_arch=$([ "${{ matrix.config.arch }}" = "amd64" ] && echo x86_64 || echo aarch64)
         sudo cp -P vendor/cuobj/lib/${cuobj_arch}/* /usr/local/lib/
+        # libminiocpp is static; its vcpkg archives must sit beside it to link.
+        sudo cp -P vcpkg_installed/*/lib/*.a /usr/local/lib/
         sudo ldconfig
 
     - name: Set up Go
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -90,6 +90,10 @@ jobs:
       env:
         CGO_ENABLED: "1"
       run: |
+        # libminiocpp is static, so minio-go's own -lminiocpp no longer drags in
+        # the C++ runtime or the vcpkg archives; name them from the one list the
+        # build script and goreleaser config also use.
+        export CGO_LDFLAGS="-L/usr/local/lib $(cat scripts/rdma-cgo-libs.txt)"
         go build -tags=rdma -o /tmp/warp .
         go vet -tags=rdma ./...
         go test -race -tags=rdma ./...
--- a/.github/workflows/qreleaser-test.yml
+++ b/.github/workflows/qreleaser-test.yml
@@ -71,13 +71,15 @@ jobs:
           ../vcpkg/vcpkg install
           cmake . -B ./build \
             -DCMAKE_BUILD_TYPE=Release \
-            -DBUILD_SHARED_LIBS=ON \
+            -DBUILD_SHARED_LIBS=OFF \
             -DCMAKE_INSTALL_PREFIX=/usr/local \
             -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake \
             -DMINIO_CPP_ENABLE_RDMA:BOOL=ON
           cmake --build ./build --config Release -j 4
           sudo cmake --install ./build
           sudo cp -P vendor/cuobj/lib/x86_64/* /usr/local/lib/
+          # libminiocpp is static; its vcpkg archives must sit beside it to link.
+          sudo cp -P vcpkg_installed/*/lib/*.a /usr/local/lib/
           sudo ldconfig
           # The libraries are installed system-wide now, and the workspace is
           # what qreleaser releases from, so leave no extra trees behind in it.
--- a/.github/workflows/qreleaser-test.yml
+++ b/.github/workflows/qreleaser-test.yml
@@ -142,20 +142,21 @@ jobs:
         run: |
           find ${{ github.workspace }}/warp-release -type f
           "${ARCH_DIR}/warp" --version
-          # The warp-rdma variant must land beside the stock binary rather than
+          # The RDMA flavor must land beside the stock binary rather than
           # replacing it, and must actually run.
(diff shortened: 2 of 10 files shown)


harshavardhana:
## Fixes Applied Successfully

Fixed 3 file(s) based on 3 CodeRabbit feedback item(s): 2 applied, 1 deferred.

**Files modified:**
- `.goreleaser/qreleaser.yaml`
- `.qreleaser.yml`
- `.github/workflows/qreleaser-test.yml`

**Commit:** `fcdf61b`

**Preserve inherited `CGO_CFLAGS` and `CGO_LDFLAGS`** — applied. I verified the premise rather than taking it from the review: with `CGO_CFLAGS=-I/from-process` exported and `-I/from-config` in a build's `env:`, the build sees only `-I/from-config`. Inherited flags are now carried through.

**Add a runtime search path for `MINIOCPP_PREFIX`** — applied, with one deliberate change. The rpath is worth having, so a non-default prefix still resolves. But the suggested `--disable-new-dtags` emits `DT_RPATH`, which outranks `LD_LIBRARY_PATH` and would stop a user of the published binary pointing at their own libminiocpp. Since this artifact is distributed to hosts that supply their own stack, it now emits `DT_RUNPATH` instead, confirmed with `readelf`. (`DT_RPATH` remains correct for the self-contained tarball from `scripts/build-rdma.sh`, where the libraries are bundled and cuObj must resolve transitively.)

**Pin every action to an immutable commit SHA** — deferred, consistent with the same decision on #500. All six warp workflows use floating tags; pinning one file would be inconsistent, and it is better done repo-wide in its own PR.

### Also in this commit: the release-test failure

Separate from the review, `release-test` had got past the build and failed in the transform with `Binary not found: dist/warp-rdma_linux_amd64_v1/warp`. That turned out to be qreleaser telling us it already has an RDMA convention — `-rdma` is a flavor like `-fips`, expecting the plain binary name inside the dist directory and suffixing the published files. This PR now follows it and drops the `variants` declaration, so the artifact is `warp.rdma` and nothing beyond the current qreleaser release is required.

harshavardhana:
## Fixes Applied Successfully

Fixed 2 file(s) based on 2 CodeRabbit feedback item(s), both applied.

**Files modified:**
- `.goreleaser/qreleaser.yaml`
- `.github/workflows/qreleaser-test.yml`

**Commit:** `13b92b5`

**Keep indirect `cuObj` dependencies loadable with a custom `MINIOCPP_PREFIX`.** Both halves of this check out, and I verified them rather than reasoning about them:

- `libminiocpp.so.0.4.0` carries no `RUNPATH` or `RPATH` of its own, so its cuObj dependencies resolve from the host's library search path regardless of what the executable declares. `DT_RUNPATH` being non-transitive therefore does mean a custom prefix could load libminiocpp and still fail on `libcuobjclient`.
- `-Wl,-rpath` alone leaves the tag to the host linker's default. That is exactly why `scripts/build-rdma.sh` needed an explicit `--disable-new-dtags` to get `DT_RPATH` for the bundled archive.

So the tag is now pinned with `--enable-new-dtags`, confirmed with `readelf`. `DT_RUNPATH` remains the deliberate choice here: `DT_RPATH` would outrank `LD_LIBRARY_PATH` and stop a user of the published binary substituting their own libminiocpp. What the rpath cannot deliver is now documented instead of implied — a non-default prefix has to be on the host's search path anyway, which is what the test workflow does with `ldconfig`.

**Assert the expected RDMA CLI errors.** Correct as reported: the flavor's exit status was discarded and the check only required a string to be absent, so a crash or any unrelated failure would have passed. The stock binary must now refuse for the documented reason, and the flavor must get past the RDMA gate and fail at the network layer. Both assertions were run against real binaries from a transformed release layout before pushing.

### CI

`release-test` passed on the previous commit at 13m52s — the first green run of the full RDMA release path: libminiocpp provisioned, both builds produced, the transform placing `warp.rdma`, and the verify assertions holding on a real runner.

PR #502, opened 2026-08-11
docs: add an RDMA guide, and give the RDMA packages checksums

harshavardhana:
Follow-up to #501. Two things: a guide for people standing up an RDMA client, and a fix for a gap I found while verifying that PR's release artifacts.

## RDMA.md

The RDMA material outgrew the README, so it moves to `RDMA.md`, written in the order someone actually works: install the package or archive, configure cuObj, run a benchmark, read the errors when it does not work. The README keeps a short summary and a link.

The part nothing documented today is **`cuobj.json`**. Warp does not install it, and without one naming the client NIC every transfer quietly falls back to HTTP — which, given the fallback is silent, is the failure this guide most needs to prevent.

The settings and the shape of the example come from clients that are actually working rather than from the vendored sample: pin a single NIC, multipath off, logging at `ERROR` rather than the sample's `TRACE`, and `use_pci_p2pdma` with `rdma_peer_type: dmabuf` for GPU-Direct. The sample's `<client-nic-ip>` placeholder is called out, since copying it unedited is the easiest way to end up on HTTP without noticing.

Troubleshooting covers the errors this work actually produced, including the CUDA runtime/driver mismatch (`the oldest CUDA runtime on this host is X, but the NVIDIA driver only supports up to Y`) and the loader error when the cuObj libraries are not found.

## The RDMA packages had no checksums

Verifying #501's `release-test` artifact turned this up:

```
warp_10.0.0_amd64.deb           warp_10.0.0_amd64.deb.sha256sum
warp-rdma_10.0.0_amd64.deb      (none)
```

pkger writes checksums against the names it chose, and the post-transform hook renames the packages to `warp-rdma_*` on the way into the layout, which invalidates them. They were being dropped, so the RDMA packages shipped unchecksummed while the stock ones did not. The hook now regenerates them after the rename; verified with `sha256sum -c`.

## Formatting

A second commit runs prettier over the yaml and markdown. It is kept separate so the content is reviewable on its own. I also repaired a stray terminal escape sequence that had been written into `.gitignore`, which would have broken the `*.exe` rule, and added `warp-release/` to it since a local hook run creates that directory.

## Verification

`go build`, `golangci-lint`, `bash -n`, `goreleaser check`, YAML parse of every workflow, and `prettier --check` all pass. The package installs and runs — I installed the deb on a GPU host and confirmed it resolves its bundled cuObj libraries from `/usr/lib/warp` with no `LD_LIBRARY_PATH` and accepts `--rdma=cpu`.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive S3 over RDMA documentation covering setup, configuration, runtime usage, distributed operation, and troubleshooting.
  - Expanded guidance for RDMA, benchmarks, analysis, profiling, metrics, and Iceberg REST catalog benchmarks.

- **Bug Fixes**
  - Release packages now receive SHA-256 checksums after RDMA package renaming.
  - Release validation verifies checksum files and package integrity.

- **Documentation**
  - Improved formatting and clarity across project, RDMA, benchmark, and release documentation.

- **Chores**
  - Normalized workflow and release configuration formatting.
  - Excluded release output from version control.

### DIFF
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -3,10 +3,10 @@ name: Build (RDMA)
 on:
   pull_request:
     branches:
-    - master
+      - master
   push:
     branches:
-    - master
+      - master
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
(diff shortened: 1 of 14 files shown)


harshavardhana:
## Fixes Applied Successfully

Fixed 3 file(s) based on 11 CodeRabbit feedback item(s): 8 applied, 3 deferred.

**Files modified:**
- `RDMA.md`
- `README.md`
- `.github/workflows/qreleaser-test.yml`

**Commit:** `aeba5c2`

The applied items were all places where the guide asserted something it could not back up:

- **Verify the APK before bypassing package trust** — the install steps used `--allow-untrusted` without verifying anything, in the very PR that adds the RDMA checksums. The guide now checks the `.sha256sum` first and explains why apk needs the flag.
- **Document the runtime library path for the standalone build** — the `go build` snippet produced a binary that loads the cuObj libraries at run time with nothing pointing at them. Three ways to resolve it are now given.
- **Document the base C/C++ runtime requirement** — "the only things it needs from the host" omitted `libc` and `libstdc++`, which the binary links dynamically.
- **Include CMake 3.31 or newer in the source-build prerequisites** — the version was named in prose but missing from the install command, and distribution packages are usually older than `build-rdma.sh` accepts.
- **Describe RDMA as an attempt, not a guarantee** — correct, and it matters here: the HTTP fallback is silent, which the guide warns about a few lines earlier.
- **Add a checksum verification to the release test** — `release-test` now runs `sha256sum -c` over the RDMA package checksums. Worth having because the hook regenerates them after renaming, so a stale one would otherwise pass unnoticed. Exercised against a real hook run before pushing.

Two were pre-existing defects in files this PR already touches, both confirmed against master rather than assumed to be reflow damage: the **ZIP section** broke off mid-sentence ("...extension that allows"), and **`--analyze.dur`** was written with a single dash where the rest of the README uses two. I checked the flag name itself is valid (`cli/analyze.go:50`), so that one was purely a consistency nit.

Deferred, with reasons:

- **Pin external CI actions and tool versions** — consistent with the same decision on #500 and #501. It is a repo-wide policy change better made in its own PR covering all six workflows than piecemeal in a docs PR.
- **Add language tags to fenced code blocks** — the four fences are pre-existing README style, untouched by this PR beyond reflow.
- **Add a concurrency policy for repeated CI runs** — `go.yml` is unrelated to this PR; it appears in the diff only because prettier reformatted it.

harshavardhana:
## Fixes Applied Successfully

Fixed 2 file(s) based on 2 CodeRabbit feedback item(s), both applied.

**Files modified:**
- `.github/workflows/qreleaser-test.yml`
- `README.md`

**Commit:** `bc9db30`

**Verify one checksum manifest for every RDMA package.** Correct, and a fair hit on the check I added in the previous round. I reproduced it before fixing: with two packages and one checksum file, the old check passed, because it iterated the manifests that exist rather than requiring one per package — precisely the failure this PR set out to catch. It now walks the packages, lets `sha256sum -c` fail on a missing manifest, and fails outright when no RDMA packages were produced at all.

Verified both ways: the tightened check errors on a deliberately missing checksum, and still passes against a real hook run over all three package types.

**Use a hyphen in the compound adjective.** Applied, on the line I had already edited for `--analyze.dur`.

PR #503, opened 2026-08-11
build(rdma): provision the release host and publish warp-rdma for arm64

harshavardhana:
Two related changes to the S3-over-RDMA release path: make the release host's toolchain reproducible, and publish `warp-rdma` for arm64 as well as amd64.

## 1. Release-host provisioning

The `warp-rdma` build links libminiocpp through cgo, so the C++ toolchain has to exist on the release host — the qreleaser checkout is wiped per run, so it cannot be provisioned from the build. `.goreleaser/qreleaser.yaml` documented that requirement in a comment, but nothing installed or verified it.

That gap broke the v1.6.0 release: the host carried a pre-`c_api.h` libminiocpp (v0.3.0 era, installed for something else), and the release died 20 minutes in with

```
rdma.go:17:11: fatal error: miniocpp/c_api.h: No such file or directory
```

Two more failures were queued behind it — `/usr/local/lib` had 2 of the 9 archives `scripts/rdma-cgo-libs.txt` names, and none of the cuObj shared objects `scripts/release-post-transform.sh` copies out of the prefix under `set -euo pipefail`.

`scripts/setup-rdma-release-host.sh` provisions and verifies a host: pinned minio-cpp and vcpkg refs in lockstep with `build-rdma.sh` and `go-rdma.yml`, enforced on every run; an existing install moved aside to a timestamped backup rather than overwritten; a new-enough CMake kept in its own work directory instead of upgrading a package other builds on a shared host depend on. `--verify-only` asserts the prefix can satisfy the exact link line qreleaser composes, which is cheap enough to use as a release preflight and would have turned this into a 2-second failure.

## 2. arm64

The RDMA build needs the C++ side built for every architecture it publishes, so arm64 is cross-built on the amd64 release host — which already had the aarch64 toolchain and arm64 RDMA libraries provisioned for the minio server's RDMA build.

- `scripts/rdma-cross/` — cmake toolchain and vcpkg overlay triplet. minio-cpp already selects its vendored cuObj libraries from `CMAKE_SYSTEM_PROCESSOR`, so the aarch64 ones resolve without patching it.
- one prefix per architecture: amd64 natively at the prefix, arm64 at `<prefix>/aarch64-linux-gnu`. Verification uses `readelf` to confirm a prefix and a smoke build really target the architecture they claim — the check that separates a working cross prefix from one holding host archives.
- `qreleaser.yaml` gains arm64 through a goreleaser `overrides` entry rather than `{{ .Arch }}` templating, so the amd64 link line stays byte for byte what it was.
- `release-post-transform.sh` packages the cuObj libraries from the prefix matching the architecture being packaged.

## A live bug this turned up

Enforcing static linkage surfaced a defect predating this PR. A `libcurlpp.so` dated Nov 2024, left in `/usr/local/lib` by an unrelated install, shadowed the `libcurlpp.a` beside it — `ld` prefers a shared object over an archive in the same `-L`. The amd64 `warp.rdma` therefore carried `NEEDED libcurlpp.so.1`, a runtime dependency the package neither bundles nor declares, and would not have started on a host without libcurlpp. arm64 was unaffected; its prefix was clean.

Shadowing shared objects are now moved aside with stale headers, `verify_prefix` rejects them, and the smoke build fails if the binary picks one up dynamically.

## Testing

On the actual release host:

- both prefixes provisioned and verified; `readelf` confirms AArch64 archives in the arm64 prefix, X86-64 in the amd64 one
- `goreleaser build --id warp-rdma` produces both binaries; the arm64 one is `ELF 64-bit LSB executable, ARM aarch64` with `RUNPATH [/usr/lib/warp]`
- after the shadowing fix both binaries need exactly `libcuobjclient.so.1`, `libcufile.so.0` and the base C/C++ runtime
- `release-post-transform.sh` packages both arches; unpacking the arm64 deb confirms the binary *and* the bundled `libcuobjclient.so.1.2.0` are AArch64
- reruns are idempotent: unchanged refs leave build directories intact, a moved ref is detected and its stale artifacts dropped, and an install identical to what is already there is left in place

## Summary by CodeRabbit

* **New Features**
  * Added automated Linux host setup for RDMA-enabled release builds.
  * Added configurable installation paths, cross-compilation, verification checks, and optional smoke testing.
  * Added architecture-aware tooling, libraries, and packaging for Linux amd64 and arm64.
  * Safely preserves existing installations during setup.

* **Documentation**
  * Updated build and installation guidance with arm64 requirements and package names.
  * Documented separate native and cross-compilation configurations and behavior on unprovisioned hosts.

* **Bug Fixes**
  * Improved release packaging to select the correct architecture-specific libraries and dependencies.

### DIFF
--- a/.github/workflows/qreleaser-test.yml
+++ b/.github/workflows/qreleaser-test.yml
@@ -35,61 +35,28 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
 
-      # The release builds the warp-rdma variant, which links libminiocpp
-      # through cgo, so the runner needs the same RDMA prerequisites as a real
-      # release host. Without them the pipeline fails on miniocpp/c_api.h.
-      - name: Checkout minio-cpp
-        uses: actions/checkout@v4
-        with:
-          repository: minio/minio-cpp
-          path: "minio-cpp"
-          persist-credentials: false
-
-      # Pinned: vcpkg's port scripts track the newest CMake, and a floating
-      # checkout breaks against whatever CMake the runner image ships.
-      - name: Checkout vcpkg
-        uses: actions/checkout@v4
-        with:
-          repository: microsoft/vcpkg
-          ref: "2026.07.29"
-          path: "vcpkg"
-          persist-credentials: false
-
-      - name: Install RDMA dependencies
-        run: |
-          sudo apt-get -qy update
-          sudo apt-get -qy install libibverbs-dev librdmacm-dev libnuma-dev
-
       - name: Set up CMake
         uses: lukka/get-cmake@v4.4.2
 
-      - name: Build and install libminiocpp.so with RDMA
-        shell: bash
-        run: |
-          ./vcpkg/bootstrap-vcpkg.sh -disableMetrics
-          cd minio-cpp
-          ../vcpkg/vcpkg install
-          cmake . -B ./build \
-            -DCMAKE_BUILD_TYPE=Release \
-            -DBUILD_SHARED_LIBS=OFF \
-            -DCMAKE_INSTALL_PREFIX=/usr/local \
-            -DCMAKE_TOOLCHAIN_FILE=../vcpkg/scripts/buildsystems/vcpkg.cmake \
-            -DMINIO_CPP_ENABLE_RDMA:BOOL=ON
-          cmake --build ./build --config Release -j 4
-          sudo cmake --install ./build
-          sudo cp -P vendor/cuobj/lib/x86_64/* /usr/local/lib/
-          # libminiocpp is static; its vcpkg archives must sit beside it to link.
-          sudo cp -P vcpkg_installed/*/lib/*.a /usr/local/lib/
-          sudo ldconfig
-          # The libraries are installed system-wide now, and the workspace is
-          # what qreleaser releases from, so leave no extra trees behind in it.
-          cd "${GITHUB_WORKSPACE}" && rm -rf minio-cpp vcpkg
-
+      # Go before the prefixes: the provisioning script link-tests each
+      # architecture with a smoke build when a toolchain is available.
       - name: Setup Go
         uses: actions/setup-go@v6
         with:
           go-version-file: "go.mod"
           cache: false
+
+      # The release builds warp-rdma for linux/amd64 and linux/arm64, and both
+      # link libminiocpp through cgo, so the runner needs the same prefixes as a
+      # real release host. Without them the pipeline fails on miniocpp/c_api.h,
+      # or on the missing aarch64 compiler.
+      #
+      # Running the release-host script rather than repeating its steps keeps CI
+      # from drifting away from how the host is actually provisioned, and covers
+      # the arm64 cross build on every PR. It keeps its checkouts outside the
+      # workspace, which is what qreleaser releases from.
+      - name: Provision the RDMA prefixes
+        run: ./scripts/setup-rdma-release-host.sh
 
       - name: Configure git
         run: |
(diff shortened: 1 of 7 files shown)


harshavardhana:
## Fixes Applied Successfully

Fixed 1 file based on 3 CodeRabbit feedback item(s).

**Files modified:**

**Commit:** `c39ee88`

Two of the three were applied with a variation on the suggested fix:

- **apt privileges** — routing apt through `as_root` verbatim would have run `apt-get install` unprivileged whenever a custom `--prefix` is writable by the invoking user, since that helper gates on prefix writability rather than on being root. Added a separate `run_privileged` helper that gates on `id -u`, used it for both apt calls, and had `as_root` delegate to it. Fixes the root-without-sudo case the review identified.
- **pinned refs** — applied `sync_checkout`, plus wiping `build/` and `vcpkg_installed/` when the minio-cpp ref actually moves; otherwise the install step's `cp vcpkg_installed/*/lib/*.a` copies the previous ref's archives into the prefix even with the checkout corrected. Cloning also happens into a cleaned directory, so a leftover partial clone doesn't break the clone branch.
- **smoke build** — the false-confidence concern was valid, but a hard `exit 1` would have failed the run that provisioned the current release host, which was executed from a copied `scripts/` directory with no `go.mod` alongside. Kept the guard, added a stderr WARNING, and made the closing line state whether the link test actually ran.

**Validation:** `bash -n` and `shfmt -d` clean; `--verify-only` passes on the provisioned release host and correctly reports every missing library on an unprovisioned one; both `sync_checkout` branches exercised against real checkouts (unchanged ref leaves the build dirs intact, moved ref is detected).

The latest autofix changes are on the `rdma-release-host-setup` branch.

harshavardhana:
## arm64 support for `warp-rdma` (`b5aed3d`)

The RDMA build links libminiocpp through cgo, so a second published architecture needs the C++ side built for it on the release host. Done by cross-building on the amd64 release host, which already had the aarch64 toolchain and arm64 RDMA libraries provisioned for the minio server's RDMA build.

- `scripts/rdma-cross/` — cmake toolchain + vcpkg overlay triplet. minio-cpp already picks its vendored cuObj libraries from `CMAKE_SYSTEM_PROCESSOR` (`CMakeLists.txt:70-79`), so the aarch64 ones resolve without patching it.
- `setup-rdma-release-host.sh` — one prefix per architecture: amd64 natively at the prefix, arm64 cross-built at `<prefix>/aarch64-linux-gnu`. Verification is per architecture and now uses `readelf` to confirm a prefix and a smoke build really target the architecture they claim, which is what distinguishes a working cross prefix from one holding host archives. `--no-cross` provisions the host architecture only.
- `qreleaser.yaml` — arm64 via a goreleaser `overrides` entry rather than `{{ .Arch }}` templating, so the amd64 link line stays byte for byte what it was.
- `release-post-transform.sh` — packages the cuObj libraries from the prefix matching the architecture being packaged.

### A live bug this turned up in the amd64 build

Enforcing static linkage surfaced a defect that predates this change. A `libcurlpp.so` dated Nov 2024, left in `/usr/local/lib` by an unrelated install, shadowed the `libcurlpp.a` installed beside it — `ld` prefers a shared object over an archive in the same `-L`. So `warp.rdma` for amd64 carried `NEEDED libcurlpp.so.1`, a runtime dependency the package neither bundles nor declares in `pkg-scripts/rdma-deps.yaml`; it would have failed to start on a host without libcurlpp. arm64 was unaffected, since its prefix was clean.

Shared objects shadowing an archive are now moved aside along with stale headers, `verify_prefix` rejects them, and the smoke build fails if the binary picks any of them up dynamically.

### Validation on the real release host

- both prefixes provisioned and verified; `readelf` confirms `AArch64` archives in the arm64 prefix and `X86-64` in the amd64 one
- `goreleaser build --id warp-rdma` produces both binaries; the arm64 one is `ELF 64-bit LSB executable, ARM aarch64` with `RUNPATH [/usr/lib/warp]`
- after the shadowing fix, both binaries need exactly `libcuobjclient.so.1`, `libcufile.so.0` and the base C/C++ runtime — no `libcurlpp`
- `release-post-transform.sh` packages both arches: `warp-rdma_<v>_arm64.deb`, `warp-rdma-<v>-1.aarch64.rpm`, `warp-rdma_<v>_aarch64.apk`. Unpacking the arm64 deb confirms both the binary and the bundled `libcuobjclient.so.1.2.0` are AArch64 — the check that proves the per-arch prefix wiring, since a shared prefix would have put x86-64 libraries in the arm64 package.
- the aarch64 cuObj libraries carry the same versions as x86_64 (`1.18.0`, `1.2.0`), so `pkg-scripts/rdma-contents.yaml` needed no change

### Two things left alone, deliberately

1. **`pkg-scripts/rdma-deps.yaml` declares no C++ runtime.** Both binaries need `libstdc++.so.6`, `libgcc_s.so.1` and `libc.so.6`, but the deps list only covers the RDMA stack. In practice these are present everywhere, so this is latent rather than broken — but it is the same class as the libcurlpp bug, and adding `libstdc++6`/`libstdc++` would close it. Not touched here since it changes package metadata beyond the arm64 scope.
2. **No CI job for the cross build.** `go-rdma.yml` already builds arm64 natively, so the cross path's unique risk is the vcpkg cross build — a ~20 minute job on every PR touching these files. Validated on the release host instead.

harshavardhana:
## Fixes Applied Successfully

Fixed 4 file(s) based on 8 CodeRabbit feedback item(s).

**Files modified:**
- `.goreleaser/qreleaser.yaml`
- `scripts/rdma-cross/aarch64-linux-gnu.cmake`
- `RDMA.md`

**Commit:** `473e430`

All eight were valid. Three were applied with a variation on the suggestion:

- **reset the existing checkout** — applied the `git reset --hard` half, which fixes the stated failure (a modified tracked file makes `checkout --detach` refuse to run). Did not clean untracked paths: `build/` and `vcpkg_installed/` are untracked, and they are exactly what makes a rerun incremental. The script already drops them deliberately when the pinned ref moves, so an allowlist-based clean would add risk without adding coverage.
- **amd64-only release host** — the sharpest finding here: `skip` covers the whole build id and goreleaser has no per-architecture equivalent, so a `--no-cross` host now fails the release rather than publishing amd64. Documented in `qreleaser.yaml` and warned about from the script when `--no-cross` is used, rather than splitting arm64 into a second build id — that id could not be named `warp-rdma`, and qreleaser derives the published flavor from the id name, so it would risk the `warp.rdma` naming this PR depends on.
- **ports repository** — applied the `ID`-is-ubuntu gate, which prevents Ubuntu suites being written into a Debian host's persistent apt config. Skipped the secondary suggestion to add `[arch=amd64]` qualifiers to the host's existing `sources.list`: that edits unrelated system configuration, and `apt-get update || true` already tolerates the resulting 404s.

The `--verify-only` remediation message was my own error from the previous autofix round — that flag returns before the smoke build, so it cannot link-test anything. It now points at a plain rerun and states that link compatibility is unverified when the smoke build is skipped.

`CXX` in the smoke build is inert today, since these cgo packages compile C rather than C++. Fixed anyway: reproducing the release environment exactly is the point of that step, and the divergence would go unnoticed if a C++ translation unit ever appeared.

**Validation:** `bash -n` and `shfmt -d` clean; `--help` now stops at the end of the comment block; a full provisioning run on the release host exits 0 with both prefixes verified and both smoke builds passing. The reset path was tested directly by modifying a tracked file in the arm64 checkout first — the run recovered and left the checkout clean, where previously it would have failed at `checkout --detach`.

The latest autofix changes are on the `rdma-release-host-setup` branch.

PR #508, opened 2026-08-15
feat(rdma): pre-allocate the pinned buffer pool, and carry objects past 4 GiB

harshavardhana:
Two problems with one cause: RDMA buffers were allocated **inside the timed loop** and sized to the object.

## Pay the cost before the run, not during it

The first operation of every worker paid for a page-aligned allocation and the `ibv_reg_mr` that pins the pages — and paid again whenever a larger object turned up and the buffer had to grow. None of that is the transfer being measured.

A pool of one buffer per worker is now allocated in `Prepare`, so the cost lands before the benchmark starts.

## Objects larger than a descriptor

Sizing the buffer to the object also meant anything past the **4 GiB** an RDMA descriptor can name asked for that much pinned memory and *then* failed on every single operation:

```
warp: <ERROR> upload error: RDMA put: buffer size 5368709120 must be between 0
      and the 4294967295 bytes an RDMA descriptor can describe
```

warp pinned 5 GiB per worker to be told it was too big. Above the threshold the upload now streams as multipart through a fixed 64 MiB buffer, which minio-go turns into RDMA `UploadPart`s ([minio-go#2284](https://github.com/minio/minio-go/pull/2284)), so pinned memory is a window rather than the object.

## The buffer-provided case is unchanged

An object that fits a descriptor still goes out as a **single RDMA transfer from the caller's buffer** — no allocation, no chunking. That remains the fast path and this PR does not disturb it.

`--rdma.window` forces streaming below the threshold, for measuring the streaming path against objects that would otherwise go single-shot.

## Measured

Two-rail mlx5 host, live 4-node cluster:

| | |
|---|---|
| 4 MiB objects, 16 workers | **7177 MiB/s** — unchanged by the pool |
| 5 GiB objects | **upload cleanly**, where they previously failed on every operation |

A server-side trace confirms the small-object path stays single-shot: 13238 `s3.PutObject`, zero `s3.PutObjectPart`.

## Note on chunk size

I tested larger RDMA chunks, expecting them to help. They do not — 16 MiB beats both:

| chunk | throughput |
|---|---|
| 16 MiB | 325.57 MiB/s |
| 256 MiB | 280.87 MiB/s |
| 1 GiB | 211.26 MiB/s |

The fill is serial, so a 1 GiB part spends seconds being read before a byte can move and the pipeline cannot hide it. Chunking stays at libminiocpp's 16 MiB default and this PR adds no knob for it.

## Summary by CodeRabbit

* **New Features**
  * Added configurable RDMA transfer windows through a command-line option and environment variable.
  * Enabled multipart RDMA uploads for objects larger than 4 GiB.
  * Improved RDMA transfer efficiency with preallocated, reusable buffers.
  * Added automatic handling for transfers exceeding RDMA descriptor limits.
  * Improved buffer sizing for large-object transfers.
* **Bug Fixes**
  * Invalid RDMA window values now produce a clear configuration error.

### DIFF
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -42,7 +42,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           repository: minio/minio-cpp
-          ref: f89cc4d25d30057f30a0c3adabe2244c9e222486
+          ref: edba72c940b5372afb60817bb5a14124b572135b
           path: "minio-cpp"
           persist-credentials: false
 
--- a/.github/workflows/go-rdma.yml
+++ b/.github/workflows/go-rdma.yml
@@ -36,13 +36,12 @@ jobs:
       # Pinned in lockstep with scripts/build-rdma.sh and
       # scripts/setup-rdma-release-host.sh: this job copies vendor/s3rdma out of
       # the checkout, so a floating main makes the build race whatever landed
-      # there. A commit rather than a tag -- the RDMA transport moved to
-      # libs3rdma after v0.5.0 and no release carries it yet.
+      # there. v0.6.0 is the first release carrying the libs3rdma transport.
       - name: Checkout minio-cpp
         uses: actions/checkout@v5
         with:
           repository: minio/minio-cpp
-          ref: edba72c940b5372afb60817bb5a14124b572135b
+          ref: v0.6.0
           path: "minio-cpp"
           persist-credentials: false
 
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -189,6 +189,12 @@ var ioFlags = []cli.Flag{
 		Value:  "",
 	},
 	cli.StringFlag{
+		Name:   "rdma.window",
+		Usage:  "Stage RDMA PUTs through a pinned window of this size instead of pinning the whole object, e.g. \"64MiB\". Objects upload as multipart, so the ETag gains a -N suffix. Empty pins the whole object.",
+		EnvVar: appNameUC + "_RDMA_WINDOW",
+		Value:  "",
+	},
+	cli.StringFlag{
 		Name:   "region",
 		Usage:  "Specify a custom region",
 		EnvVar: appNameUC + "_REGION",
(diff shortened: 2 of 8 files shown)


harshavardhana:
## Fixes Applied Successfully

Fixed 3 file(s) based on 7 CodeRabbit feedback item(s) — 5 applied, 2 declined.

**Files modified:**
- `pkg/bench/put.go`
- `cli/flags.go`

**Commit:** `39f8dd8`

### Applied

- **Release replacement RDMA buffers.** The real one. Moving to a pool dropped the `defer` that released the per-worker buffer, and `Cleanup` walks only the pool — so a buffer allocated because an object exceeded the pool's size stayed pinned for the life of the process. Ownership is now tracked so the replacement is freed at worker exit and the pooled one is left to `Cleanup`, rather than freed twice.
- **Pass the worker context to multipart PUT.** All three `PutObject` call sites, not just the new one. Worth flagging that this changes pre-existing behaviour: `nonTerm` was deliberate (`// Non-terminating context.`), so an operation cancelled at shutdown is now reported as an error where it previously ran to completion. The trade is that Ctrl-C no longer waits out a multi-GiB upload.
- **Free `u.rdmaPool` when bucket creation fails** — `Prepare` allocates before `createEmptyBucket`, and a failed `Prepare` returns without the runner calling `Cleanup`.
- **Reject `--rdma.window` above `MaxInt64`** rather than wrapping negative, which reads as "not set" and silently ignores the request.
- **Update the stale checkout comment** — a second one in `sync_checkout` still called the pin a commit.

### Declined

- **Pin `minio-cpp` to one immutable revision.** This asks to revert `v0.6.0` to `edba72c`. A tag is mutable and a commit is not, so the instinct is sound, but moving to the tag was a deliberate call once v0.6.0 shipped the libs3rdma transport — and `MINIO_CPP_REF` still selects a commit when one is wanted.
- **Pin `actions/checkout` to a full commit SHA.** This repo uses `@v5` across every workflow; pinning the one line touched here leaves the file less consistent, not more. Worth doing repo-wide as its own change.

### Verified

`go vet` clean, `go test -race ./pkg/bench/` passes, `golangci-lint` 0 issues, and `-tags=rdma` builds and vets on a two-rail mlx5 host. Both paths re-run live against a 4-node cluster afterwards: 4 MiB objects at 6790 MiB/s single-shot, and 6 GiB objects streaming at 658 MiB/s with no errors.

The latest autofix changes are on the `feat/rdma-stream` branch.

PR #512, opened 2026-08-16
perf(rdma): stage GPU uploads through a reusable pinned host buffer

harshavardhana:
`stageToGPU` allocated a Go-heap bounce buffer on every operation and `cudaMemcpy`'d out of it:

```go
host := make([]byte, n)                     // 64 MiB Go-heap alloc, every op
io.ReadFull(src, host)
C.warp_cuda_memcpy_h2d(b.ptr, &host[0], n)  // from PAGEABLE memory
```

Pageable memory cannot DMA, so the CUDA runtime copies through its own internal pinned staging
buffers first — a third copy on top of the two warp already makes, plus a 64 MiB allocation per
operation for the GC to chase.

### Measurement

`warp put`, 4-node AIStor cluster, `--concurrent=8`, 20s, H200 host with two 400 Gb RoCE rails.
Taking the `--rdma=gpu` minus `--rdma=cpu` request time as the cost of the extra staging:

| obj size | gpu − cpu request time | implied H2D bandwidth |
|---|---|---|
| 8 MiB | 5.5 ms | 1.45 GiB/s |
| 32 MiB | 18.5 ms | 1.69 GiB/s |
| 64 MiB | 30.0 ms | 2.08 GiB/s |
| 256 MiB | 77.6 ms | 3.21 GiB/s |

A pinned H2D copy on this hardware runs 25-50 GB/s, so this path was leaving better than 10× on the
table. Throughput followed: `--rdma=gpu` sat below `--rdma=cpu` at every size tested
(3099 vs 4347 MiB/s at 8 MiB, 8187 vs 12150 MiB/s at 256 MiB).

The rising bandwidth across sizes says there is also a fixed per-operation component — the
allocation and the memcpy launch — so expect this to close most, not all, of the gap.

### Change

Allocate the staging buffer alongside the device buffer in `allocRDMAGPU`, page-locked with
`cudaMallocHost`, and reuse it for the life of the worker. The buffer pool already exists to keep
allocation and registration off the timed path (`51b1f72`); this follows it there. `cudaMallocHost`
and `cudaFreeHost` join the existing dlopen'd entry points, so no CUDA headers are needed to build
and one binary still serves both `--rdma=cpu` and `--rdma=gpu`.

Costs one extra buffer's worth of host memory per worker, page-locked. GET does not use it — the
server writes to the device pointer directly.

### Measured result

Built both master and this branch with the same toolchain and the same minio-cpp prefix, run back to
back on the H200 host, `--rdma=gpu`, `--concurrent=8`, 20s, same 4-node cluster:

| obj size | master | this branch | throughput | req avg |
|---|---|---|---|---|
| 8 MiB | 3183.84 MiB/s | **3996.19 MiB/s** | **+25.5 %** | 20.2 → 16.1 ms |
| 64 MiB | 3880.82 MiB/s | **5137.68 MiB/s** | **+32.4 %** | 135.3 → 99.4 ms |
| 256 MiB | 7631.63 MiB/s | **12446.91 MiB/s** | **+63.1 %** | 269.9 → 164.8 ms |

That closes the GPU-vs-CPU gap entirely: `--rdma=cpu` on the same host measures 4346.94 /
4943.76 / 12150.33 MiB/s at those sizes, so GPU mode now matches CPU at 8 MiB and edges ahead at
64 MiB and 256 MiB.

Request-time standard deviation drops sharply as well — 23.0 → 2.6 ms at 64 MiB, 48.8 → 8.7 ms at
256 MiB — which is the per-operation 64 MiB allocation and its GC pressure leaving the hot path.

### Testing

Untagged: `go build ./...`, `go vet ./...`, `go test -race ./...`, `golangci-lint run -j16` → 0
issues.

Tagged: `go build -tags=rdma ./pkg/bench` and `go vet -tags=rdma ./pkg/bench` clean against
libminiocpp headers. `golangci-lint --build-tags=rdma` reports one new `gocritic: dupSubExpr` on
`rc != 0`, which is the same false positive it already reports for the adjacent `cudaMalloc` call on
master — the checker misreads cgo call expressions.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved GPU data transfers by using reusable pinned host memory.
  * Reduced temporary memory allocations during GPU operations.
  * Added cleanup handling for host and device transfer buffers.
* **Bug Fixes**
  * Improved error reporting when required GPU staging memory is unavailable.
  * Preserved direct data handling for CPU buffers.

### DIFF
--- a/pkg/bench/rdmabuf.go
+++ b/pkg/bench/rdmabuf.go
@@ -39,7 +39,11 @@ const MaxRDMADescriptorSize = 1<<32 - 1
 // rdmabuf_gpu.go. Both live outside the Go heap: minio-go retains the
 // pointer across the cgo boundary, which Go memory may not do.
 type rdmaBuf struct {
-	ptr  unsafe.Pointer
+	ptr unsafe.Pointer
+	// host is the pinned staging buffer PUT reads into before the copy to
+	// device. GPU mode only, so it is used solely from rdmabuf_rdma.go and
+	// is unused without -tags=rdma.
+	host unsafe.Pointer //nolint:unused
 	size int
 	mode string
 }
@@ -69,9 +73,9 @@ func allocRDMABuf(mode string, size int) (*rdmaBuf, error) {
 
 // stageToRDMABuf reads n bytes from src into the front of the buffer, which
 // may be larger because it is reused across operations. For CPU buffers this
-// is a straight ReadFull. For GPU buffers the bytes are first read into a CPU
-// bounce buffer and then uploaded via cudaMemcpy host-to-device (see
-// rdmabuf_gpu.go).
+// is a straight ReadFull. For GPU buffers the bytes go into the buffer's pinned
+// host staging area and are then uploaded via cudaMemcpy host-to-device (see
+// rdmabuf_rdma.go).
 func stageToRDMABuf(b *rdmaBuf, src io.Reader, n int) error {
 	if b == nil || n <= 0 || src == nil {
 		return nil
--- a/pkg/bench/rdmabuf_rdma.go
+++ b/pkg/bench/rdmabuf_rdma.go
@@ -29,6 +29,8 @@ package bench
 // static warp_set_device_fn warp_set_device;
 // static warp_malloc_fn warp_malloc;
 // static warp_free_fn warp_free;
+// static warp_malloc_fn warp_malloc_host;
+// static warp_free_fn warp_free_host;
 // static warp_memcpy_fn warp_memcpy;
 // static warp_error_string_fn warp_error_string;
 //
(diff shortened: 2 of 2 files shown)


harshavardhana:
## CodeRabbit Autofix Review Complete

Reviewed 1 CodeRabbit feedback item and did not apply code changes in this run.

The item asks for an end-to-end GPU RDMA benchmark before merge. That has been run and the results are in the PR description: master and this branch built from the same toolchain and the same minio-cpp prefix, run back to back on an H200 host with `--rdma=gpu --concurrent=8 --duration=20s` against a 4-node cluster.

| obj size | master | this branch | gain |
|---|---|---|---|
| 8 MiB | 3183.84 MiB/s | 3996.19 MiB/s | +25.5 % |
| 64 MiB | 3880.82 MiB/s | 5137.68 MiB/s | +32.4 % |
| 256 MiB | 7631.63 MiB/s | 12446.91 MiB/s | +63.1 % |

Pinned-memory usage is one buffer per worker at the pool's buffer size, allocated in `allocRDMAGPU` alongside the device buffer — so `--concurrent` × object size (or × `--rdma.window` on the streaming path), page-locked for the life of the run.

No code change required.

harshavardhana:
@coderabbitai review

PR #518, opened 2026-09-10
chore: build against minio-cpp 1.0.0, which installs as libminio

harshavardhana:
Builds against minio-cpp 1.0.0, which installs the library as `libminio` with a stable soname. minio-go's cgo
directive is now `-lminio` (minio/minio-go#2302), so the pin has to move here too.

**Ref pin.** Moved in lockstep across the three places that carry it -- `.github/workflows/go-rdma.yml`,
`scripts/build-rdma.sh`, `scripts/setup-rdma-release-host.sh`. Compiling against 1.0.0 headers while linking a
0.6.0 archive fails by naming a missing symbol, not a version, so a half-moved pin is expensive to diagnose.

**Link line.** `scripts/rdma-cgo-libs.txt` is the single source of truth -- the workflow, `build-rdma.sh` and
`qreleaser.yaml` read it, and `setup-rdma-release-host.sh` derives its prefix checks from it -- so renaming
`-lminiocpp` there propagates to all four.

**Brotli.** 1.0.0 builds cpp-httplib with brotli support and 0.6.0 did not: the installed `libminio.a` carries 7
undefined `Brotli*` symbols where `libminiocpp.a` carried none, so the link fails without the archives. vcpkg
already produces them and the prefix already receives them; only the list was missing. This is the part that would
have turned up as a red `go-rdma` job rather than at review.

**Upgrade path.** `setup-rdma-release-host.sh` moves a pre-1.0.0 `libminiocpp.*` aside. Its stale sweep globs the
name it installs, so without this an already-provisioned host keeps the old archive in the same `-L` as the new one.

**libs3rdma needs no separate change.** minio-cpp vendors it, so this pin carries it from 0.3.0 to 0.3.1 for both
architectures, and the soname stays `libs3rdma.so.0` -- the name the release packaging copies out -- so
`pkg-scripts/rdma-contents.yaml` and `release-post-transform.sh` are untouched.

## How to test

Verified on a release host that still held the 0.6.0 install, so the upgrade path was exercised rather than a
clean one:

```
bash scripts/setup-rdma-release-host.sh
# >>> moving 2 stale amd64 file(s) aside to /usr/local/lib/warp-rdma-backup-<ts>
# >>> /usr/local satisfies the RDMA link line for amd64
# >>> amd64 smoke build links and targets Advanced Micro Devices X86-64
# >>> arm64 smoke build links and targets AArch64
# >>> release host provisioned and link-tested for: amd64 arm64
```

Both architectures rebuilt at 1.0.0, both smoke builds link, and the resulting binary is clean:

```
$ readelf -d warp | grep NEEDED
  libs3rdma.so.0   libstdc++.so.6   libm.so.6   libgcc_s.so.1   libc.so.6   ld-linux-x86-64.so.2
```

No `libminiocpp`, `libminio` statically linked as intended, `libs3rdma.so.0` the only non-system dependency.

## Summary by CodeRabbit

* **Chores**
  * Updated RDMA build and release tooling for the latest MinIO C++ library.
  * Improved RDMA host setup, library verification, and packaging compatibility.
  * Updated RDMA linking to use system zlib and Brotli compression libraries.
  * Standardized static library naming across RDMA build and provisioning workflows.
  * Preserved existing RDMA build and test behavior.


harshavardhana:
Force-pushed. The first push failed `Build rdma` on both arches with `cannot find -lcurlpp` / `cannot find -lcurl`, which turned out to be the more interesting half of this bump.

minio-cpp 1.0.0 changes its dependency set, not just the library name: it replaces curlpp with cpp-httplib, and takes zlib from the system on Linux instead of vcpkg. Comparing the manifests:

| | 0.6.0 | 1.0.0 |
|---|---|---|
| HTTP | `curlpp` (+curl) | `cpp-httplib[openssl]` → pulls brotli |
| zlib | all platforms | windows only |

vcpkg for 1.0.0 builds exactly `brotli{common,dec,enc}`, `crypto`, `ssl`, `inih`, `INIReader`, `pugixml` — no curl, no curlpp, no libz. So `scripts/rdma-cgo-libs.txt` had to drop `-lcurlpp -lcurl` as well as gain the brotli archives, and the prefix checks had to stop demanding a `libz.a` that is no longer produced.

Worth recording why my first local verification passed while CI failed: the release host still had `libcurl.a`, `libcurlpp.a` and `libz.a` sitting in the prefix from the 0.6.0 provisioning, so the link silently used them. A clean runner has none. That is also why the orphan sweep in `setup-rdma-release-host.sh` is unconditional rather than tied to the library having changed — an orphaned `libz.a` is the one that misleads, since it still satisfies `-lz` and makes the local link differ from the released one.

Re-verified with the orphans swept, so the prefix now holds only what 1.0.0 produces:

```
>>> /usr/local satisfies the RDMA link line for amd64
>>> /usr/local/aarch64-linux-gnu satisfies the RDMA link line for arm64
>>> amd64 smoke build links and targets Advanced Micro Devices X86-64
>>> arm64 smoke build links and targets AArch64
>>> release host provisioned and link-tested for: amd64 arm64

$ readelf -d warp | grep NEEDED
  libs3rdma.so.0  libz.so.1  libstdc++.so.6  libm.so.6  libgcc_s.so.1  libc.so.6  ld-linux-x86-64.so.2
```

`libz.so.1` is now a dynamic dependency rather than statically linked, which is the correct outcome on Linux for 1.0.0.

harshavardhana:
## Review feedback addressed

Applied 1 fix across 1 file from 1 review finding.

The stale sweep globbed `libminio.*`, which also matches the `libminio.so*` files the shared-object sweep below
collects, because `static_libs()` now yields `minio` after the rename. A prefix holding a shared object from an
older install queued the same path twice, and the second `mv` failed on an already-moved source — under `set -e`,
aborting provisioning before anything was installed.

Confirmed rather than assumed, with the two globs isolated in a harness:

- old `libminio.*` — `mv: cannot stat …libminio.so`, exit 1, install step never reached
- new `libminio.a` — 3 paths, exit 0, reaches install

Then exercised end to end on a release host: planted a `libminio.so` and forced the changed-build path so the sweep
actually ran. Provisioning completed, moved 4 stale files aside with no errors, both prefixes satisfied the link
line, and both the amd64 and arm64 smoke builds linked.

Scoped narrower than the suggested edit: only the archive is swept under the current name, leaving `libminio.so*`
to the sweep that exists for it. `libminiocpp.*` stays a glob, since `static_libs()` no longer yields `miniocpp`
and nothing else covers the pre-1.0.0 shared objects.

**Files changed:**

**Commit:** `91d956d`

**Deferred:** none.

